PR #21, opened 2014-07-10
Add reconnection support.

STRML:
Renamed a few vars to make a bigger distinction between
postgresql (the connector), pg (the connection library), and the
individual connection.

Thinking this might make more sense in `loopback-datasource-juggler`. If it were moved there, this library would still require a small rework so that `connect()` instantiates a new `pg.Client` (as they cannot be reused).

### DIFF
--- a/lib/postgresql.js
+++ b/lib/postgresql.js
@@ -1,11 +1,12 @@
 /*!
  * PostgreSQL connector for LoopBack
  */
-var postgresql = require('pg');
+var pg = require('pg.js');
 var SqlConnector = require('loopback-connector').SqlConnector;
 var util = require('util');
 var async = require('async');
 var debug = require('debug')('loopback:connector:postgresql');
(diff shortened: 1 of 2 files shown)


bajtos:
Hey @STRML, thank you for the patch. It has fell trough the cracks, sorry for that.

If the changes still make sense, can you please rebase them on top of the current master?

STRML:
I think they make sense, I've rebased on top of master.

bajtos:
@STRML thanks. I'll leave it up to @raymondfeng to review and land this patch, as I am not familiar with this code base.

raymondfeng:
@STRML Do you think https://github.com/strongloop/loopback-connector-postgresql/pull/52 will obsolete this PR? It leverages the client pool from the driver.

STRML:
Yes, #52 obsoletes this. Thanks for the pointer, I will move my code back to master.

PR #28, opened 2014-09-06
Create table if not exists

fabien:
Previously, declaring a model to explicitly use a specific table name would fail, if there were multiple models that refer to the same table. Even though the table should have been dropped by automigrate before, this was not without issues, yielding a postgresql error like: `relation "my_table" already exists`.


raymondfeng:
I'm closing this PR. Please submit a new one if you are keen.

PR #42, opened 2014-12-03
fixing input date format for postgresql

venkat-baddigam:


aol-nnov:
@venkat-baddigam not sure, if fixes anything (and if this fix is required at all)
```
aol=# select timestamp with time zone '2015-01-26T13:23:20.314Z';
        timestamptz         
----------------------------
 2015-01-26 16:23:20.314+03
(1 row)

aol=# select timestamp with time zone '2015-01-26 13:23:20.314';
        timestamptz         
----------------------------
 2015-01-26 13:23:20.314+03
(1 row)
```
mind the 'hours' value!

FYI, my postgresql version
```
aol=# select version();                                                                                                                                                                                                                                                 version                                                              
-----------------------------------------------------------------------------------------------------------------------------------
 PostgreSQL 9.4.0 on x86_64-apple-darwin14.0.0, compiled by Apple LLVM version 6.0 (clang-600.0.54) (based on LLVM 3.5svn), 64-bit
(1 row)
```

PR #49, opened 2015-01-08
Escape number literals to avoid SQL injection

raymondfeng:
/to @bajtos or @ritch

### DIFF
--- a/lib/postgresql.js
+++ b/lib/postgresql.js
@@ -496,7 +496,7 @@ PostgreSQL.prototype.toDatabase = function (prop, val) {
         return 'NULL';
       }
     }
-    return val;
+    return escape(val);
   }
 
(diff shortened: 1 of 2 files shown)


ritch:
It would be nice to refactor any SQL string concatenation into parameterized versions. This LGTM for now though.

bajtos:
+1 for using parameterized SQL commands instead of string concatenating. AFAIK, parameterized SQL queries have significantly better performance, as the same querying plan can be used for all queries independently of parameter values. For queries built manualy via string concatenation, at least some SQL servers (MS SQL, IIRC) will build a new querying plan for each statement.

Having said that, this patch LGTM and should be released ASAP.

bajtos:
I have created a GH issue to follow up on the parameterized queries: https://github.com/strongloop/loopback/issues/983

Issue #50, opened 2015-01-12
How to ONLY insert given properties and not send NULL to database.


aars:
I have a TIMESTAMP column named created in my users table. The database knows the default value for it, and I want the database to handle this.

How can I make sure that a Model.save() creates an SQL query that only inserts the properties I tell it to?

For example:

``` javascript
Access.create([{ // Extends User
  uuid: rID.uuid, // Some uuid I got from some other Model
  username: 'aaron',
  password: 'test',
  email: 'test@test.test'
}]);
```

Ends up trying to insert every column of the Model/Table.

``` sql
INSERT INTO "public"."access" ("uuid","name","username","password","role","acl","created","expired","realm","credentials","challenges","email","emailverified","verificationtoken","status","lastupdated")

Failing row contains (42, 0b81bb8b-1962-4b01-9f4c-87a0938bf94b, owner, null, aaron, $2a$10$GtfE7X4BR.eyf/AI7qwHQegYdqEaEzIFnLgMy5mqkc6JQ8bB6j3e2, {"all": "rw"}, null, null, null, null, null, test@test.test, null, null, null, null).
```

How can I prevent all these null values? I want (need) my database to handle default values. I know I can set the default values ALSO in my model definition, but I don't see why I should have to. This would also mean that a simple misconfigured timezone on an application server will mess up these Date fields, since the Default Value was decided on by my NodeJS Application and not in the database.

raymondfeng:
One option is to skip columns that don't have corresponding values set in the data object during create(). For example, if data.role === undefined, we should not insert it.

aars:
That is what I want yes! I'm not sure if you're telling me this is possible (via the create() method) or that you would implement it like this.

I tried manipulating data on beforeCreate and beforeSave, but the data those methods receive is the exact data I put into Access.create(). Should I look into the connector itself (or the Model/PersistedModel of loopback/strongloop?)

raymondfeng:
Please check the connector implementations. Patches are welcome!

aars:
Will do!

aars:
@raymondfeng Check my PR. It might be a bit invasive, but I see no reason to add a switch/option for this behavior.

0candy:
@aars Does this PR https://github.com/strongloop/loopback-connector-postgresql/pull/54 resolve this issue?

PR #51, opened 2015-01-14
[Issue #50] Removes keys with undefined values.

aars:
Removes keys with undefined values from data before creating SQL query.
This prevents sending NULL values to Postgres for undefined values,
which prevents the Postgres from handling default values.

(Postgre)SQL does support undefined values. If you want to use NULL
values you should define them as such.

### DIFF
--- a/lib/postgresql.js
+++ b/lib/postgresql.js
@@ -382,6 +382,15 @@ PostgreSQL.prototype.update =
  */
 PostgreSQL.prototype.toFields = function (model, data, forCreate) {
   var self = this;
+
+  // There is no undefined in PostgreSQL and sending NULL for every
+  // undefined field prevents the database from handling default values.
+  //
+  // If you actually want to send NULL values, use NULL values.
+  Object.keys(data).forEach(function (k) {
+    if (data[k] === undefined) delete data[k];
+  });
+
   var props = self._categorizeProperties(model, data);
   var dataIdNames = props.idsInData;
   var nonIdsInData = props.nonIdsInData;
--- a/lib/postgresql.js
+++ b/lib/postgresql.js
@@ -256,7 +256,7 @@ PostgreSQL.prototype._categorizeProperties = function(model, data) {
  */
 PostgreSQL.prototype.create = function (model, data, callback) {
   var self = this;
-	data = self.filterUndefined(data);
+  data = self.filterUndefined(data);
   var props = self._categorizeProperties(model, data);
 
   var sql = [];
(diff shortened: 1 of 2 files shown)


slnode:

aars:
I know this might be a bit invasive, since this will be used on all queries, but I see no reason to add an option/switch or other way to circumvent this behavior.

aars:
If accepted and agreed on sanity of behavior I wouldn't mind doing this for the mysql connector as well.

aars:
Running into issues now. Added WIP label.

aars:
Should be all fine now.

lib/postgresql.js
```diff
@@ -374,6 +378,24 @@ PostgreSQL.prototype.update =
...
+ *
+ * @param {Object} The model instance data
+ */
+
+PostgreSQL.prototype.filterUndefined = function (data) {
+  // There is no undefined in PostgreSQL and sending NULL for every
+  // undefined field prevents the database from handling default values.
+  //
+  // If you actually want to send NULL values, use NULL values.
+  Object.keys(data).forEach(function (k) {
```
raymondfeng: Please use plain for-loop to avoid performance penalties.

raymondfeng:
@aars Please sign the CLA - https://cla.strongloop.com/agreements/strongloop/loopback-connector-postgresql

aol-nnov:
BTW, @raymondfeng, I suppose, this issue has something to do with #44 or #45.
It seems to me, this line https://github.com/strongloop/loopback-connector-postgresql/blob/master/lib/postgresql.js#L1512 is forcing such a behaviour.

If I'm right, may be it's better to fix the root cause instead of adding one more function?
May be, it would be better to remove item from props.nonIdsInData if data[key] is undefined instead of pushing null into queryParams.

//Sorry for the braindump. I need to study the code more to make my thoughts clear ;)

aol-nnov:
I went deeper, lol.

A one line fix for this issue is in my last commit for #54. Seems, this one is no longer needed. Please, verify.

raymondfeng:
@aol-nnov Do you think this PR is still relevant?

aol-nnov:
@raymondfeng I think, no. #54 supersedes it.

0candy:
@aars Can this be closed in favor of https://github.com/strongloop/loopback-connector-postgresql/pull/54?

slnode:

slnode:

slnode:

PR #52, opened 2015-01-20
Use connection pooling from the driver

raymondfeng:
/to @ritch or @bajtos

### DIFF
--- a/lib/postgresql.js
+++ b/lib/postgresql.js
@@ -58,8 +58,8 @@ function PostgreSQL(postgresql, settings) {
   // this._models = {};
   // this.settings = settings;
   this.constructor.super_.call(this, 'postgresql', settings);
-  this.postgresql = new postgresql.Client(settings.url || settings);
-  this.connection = null;
+  this.clientConfig = settings.url || settings;
+  this.pg = postgresql;
   this.settings = settings;
   if (settings.debug) {
     debug('Settings %j', settings);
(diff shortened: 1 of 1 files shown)


raymondfeng:
@ritch or @bajtos ping.

lib/postgresql.js
```diff
@@ -132,15 +108,19 @@ PostgreSQL.prototype.executeSQL = function (sql, params, callback) {
...
+      // if(err) console.error(err);
+      if (err && self.settings.debug) {
+        self.debug(err);
+      }
+      if (self.settings.debug && data) self.debug("%j", data);
+      // console.log(err);
+      if (log) log(sql, time);
+      process.nextTick(done); // Release the pooled client
+      callback(err ? err : null, data ? data.rows : null);
+    });
```
bajtos: This is not calling `done` at all, is that intentional?

bajtos: I mean `done` will be never called, is it a problem?

bajtos:
I don't know the `postgresql` client you are using, thus I can't really asses the impacts of this change. I didn't find any obvious problem besides the missing `done` call (which may be intended).

FWIW, the patch LGTM.

lib/postgresql.js
raymondfeng: `done` is used to release the client/conn back to the pool. It's passed to `process.nextTick(done);` and scheduled to be called in next tick.

bajtos: I see it now, I did not notice that line before.

Is there any particular reason why you are releasing the connection in the next tick and immediately (sync)?

Also `releaseConnection` would be a more suitable & less confusing name, since we are already using `done` as the name for the regular callback, e.g. in unit tests.

raymondfeng: I wanted to avoid the wait of callback and check possible exceptions from callback.

bajtos: My concern is that if you have 20 clients served at the same tick, then you need to open 20 connections. While if the connection was returned immediately, there is a chance that the next client can pick up the connection returned by the previous client.

Anyways, this is most likely a theoretical issue at the moment and it can be always easily fixed later.

bajtos:
No more comments from me, feel free to merge and ship this patch.

PR #53, opened 2015-01-21
autocreate schema if it does not exist during migration/update

aol-nnov:
Hi!

My models span across several schemas in Postgresql and I've found it handy if schemas are created automatically too during migration process.


slnode:

PR #54, opened 2015-01-21
provide database column default values via Loopback model description

aol-nnov:
This PR allows to define the default value for a field by the means of database server. It is especially useful when the default value is calculated by a stored procedure.

Example usage is as follows:
simpleUsage.json
```
{
    "name": "simpleModel",
    "properties": {
        "creationDate": {
            "type": "Date",
            "postgresql": {
                "dbDefault": "now()"
            }
        }
    }
}
```
Another, a more complex usecase:
myModel.json
```
{
    "name": "myModel",
    "base": "PersistedModel",
    "options": {
        "idInjection": false,
        "postgresql": {
            "schema": "store",
            "table": "item"
        }
    },
    "properties": {
        "id": {
            "type": "String",
            "id": true,
            "postgresql": {
                "dataType": "uuid",
                "dbDefault": "uuid_generate_v4()"
            }
        }
    }
}
```

I quite agree that in many cases the same behaviour can be achieved by Loopback hooks, but still there are cases exist where default value is calculated by custom stored procedure on the DB side.

Feel free to contact me if you need more comments on this PR.

### DIFF
--- a/lib/postgresql.js
+++ b/lib/postgresql.js
@@ -1273,6 +1273,8 @@ PostgreSQL.prototype.propertySettingsSQL = function (model, propName) {
   }
   var result = self.columnDataType(model, propName);
   if (!propertyCanBeNull.call(self, model, propName)) result = result + ' NOT NULL';
+
+  result += self.columnDbDefault(model, propName);
   return result;
 };
 
@@ -1382,6 +1384,20 @@ function escape(val) {
   });
   // return "q'#"+val+"#'";
   return "'" + val + "'";
+}
+
+/*!
+ * Get the database-default value for column from given model property
+ *
+ * @param {String} model The model name
+ * @param {String} property The property name
+ * @returns {String} The column default value
+ */
+PostgreSQL.prototype.columnDbDefault = function(model, property) {
+  var columnMetadata = this.columnMetadata(model, property);
+  var colDefault = columnMetadata && columnMetadata.dbDefault;
+
+  return colDefault ? (' DEFAULT ' + columnMetadata.dbDefault): '';
 }
 
 /*!
--- a/lib/postgresql.js
+++ b/lib/postgresql.js
@@ -236,7 +236,7 @@ PostgreSQL.prototype._categorizeProperties = function(model, data) {
   });
   var props = Object.keys(this._models[model].properties);
   var nonIdsInData = Object.keys(data).filter(function(key) {
-    return props.indexOf(key) !== -1 && ids.indexOf(key) === -1;
+    return props.indexOf(key) !== -1 && ids.indexOf(key) === -1 && data[key] !== undefined;
   });
   return {
     ids: ids,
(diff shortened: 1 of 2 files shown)


slnode:

raymondfeng:
Would you please add some test cases?

aol-nnov:
@raymondfeng, no problem, but it'll take some time, because I'm not familiar with test-suite, my bad...

raymondfeng:
BTW, this feature can be possibly ported to other connectors too.

aol-nnov:
yeah, I know that :)
Many (if not all) RDBMs support function calls as column default values. At least MySQL and PostgreSQL do.

aol-nnov:
@raymondfeng, just added basic tests for this PR. If it's not enough, please advice more test-cases to be considered.

raymondfeng:
Great, maybe two more tests:

1. Invalid default for DB
2. Try to create a record and read it back to see if the DB default is effective.

aol-nnov:
Okay. Will do it tomorrow! Way too late over here now.

aol-nnov:
@raymondfeng, just added more tests.
You need to integrate #51 first to pass 'should create a record with default value' though..

aol-nnov:
this PR now longer depends on #51. All tests are passing correctly.

aol-nnov:
Well, It was a headlong verdict about @aars's fix replacement. :)
As he pointed in comment to my commit bfb3f00,  there will be inconsistences between column names and values. With small extra tweaks it is possible to find a way around.
But!
His fix does not pass loopback-database-juggler tests either (at least they fail on my computer). This is the general problem and my solution won't pass them too. And the root-cause is: if all model fields are not initialised with values, wrong sql query will be generated.

I'm not yet sure how to run only partial tests, so here is the command line that I was using:
```
aol@Andreys-MBP:~/develop/web/loopback-connector-postgresql$ DEBUG=loopback:connector:postgresql ./node_modules/.bin/mocha --require ./test/init.js node_modules/loopback-datasource-juggler/test/relations.test.js -g 'should build record on scope'
```
on your master branch I have the following debug output:
```
  loopback:connector:postgresql SQL: INSERT INTO "public"."book" ("name","type") SELECT $1,$2  RETURNING "id"
Parameters: , +2ms
```
(note two empty parameters!)

And with @aar's patch applied, this query becomes:
```
  loopback:connector:postgresql SQL: INSERT INTO "public"."book" () SELECT   RETURNING "id"
```
It seems to me, that it happens if new object is saved and it's fields are not initialised.

So, it seems, we need to invent another approach to handle undefined values.
From one side, we can't substitute undefined keys of ```data``` with ```NULL```s (to utilise database default values or for leaving columns intact), from the other side, the latter query is meaningless and erroneous as column list and values list are empty.

Please, let me know, if any of my thoughts are correct.

aol-nnov:
Seems, this is the way out. All previously failing tests passed!

PR #57, opened 2015-01-29
Create model foreign key matching type of opposite part of relation ...

aol-nnov:
(even if it has a custom field type)

Hi, I've stumbled upon something recently while defining relations between my models. Here they go:
Item:
```
{
    "name": "item",
    "base": "PersistedModel",
    "options": {
        "idInjection": false,
        "postgresql": {
            "schema": "store",
            "table": "item"
        }
    },
    "properties": {
        "item_id": {
            "type": "String",
            "id": true,
            "postgresql": {
                "dataType": "uuid",
                "dbDefault": "uuid_generate_v4()"
            }
        },
        "name": {
            "type": "String"
        }
    },
    "validations": [],
    "relations": {},
    "acls": [],
    "methods": []
}
```

Variant:
```
{
    "name": "variant",
    "base": "PersistedModel",
    "options": {
        "idInjection": false,
        "postgresql": {
            "schema": "store",
            "table": "item_variants"
        }
    },
    "properties": {
        "variant_id": {
            "type": "String",
            "id": true,
            "postgresql": {
                "dataType": "uuid",
                "dbDefault": "uuid_generate_v4()"
            }
        },
        "name": {
            "type": "String"
        }
    },
    "validations": [],
    "relations": {
        "item": {
            "type": "belongsTo",
            "model": "item",
            "foreignKey": "item_id"
        }
    },
    "acls": [],
    "methods": []
}
```
Without fix attached ```item_variants``` is as follows:
```
CREATE TABLE store.item_variants
(
  variant_id uuid NOT NULL DEFAULT uuid_generate_v4(),
  name character varying(1024),
  item_id character varying(1024),
  CONSTRAINT item_variants_pkey PRIMARY KEY (variant_id)
)
```
It is logically right, as I have item.item_id defined as String, but it has custom definition for postgresql database which is not honoured during foreign key creation.
Having this fix applied, ```item_variants``` has the following structure:
```
CREATE TABLE store.item_variants
(
  variant_id uuid NOT NULL DEFAULT uuid_generate_v4(),
  name character varying(1024),
  item_id uuid,
  CONSTRAINT item_variants_pkey PRIMARY KEY (variant_id)
)
```
which is perfectly correct!

Not sure, if it should be handled in loopback-datasource-juggler to act similarly for all connectors, so I've introduced it in postgresql connector for further discussion and review.

### DIFF
--- a/lib/postgresql.js
+++ b/lib/postgresql.js
@@ -1264,6 +1264,23 @@ function applySqlChanges(model, pendingChanges, cb) {
 }
 
 /*!
+ * Define foreign key matching to foreign model key type
+ */
+PostgreSQL.prototype.defineForeignKey = function (className, key, foreignClassName, cb) {
+  var foreignMeta = this.columnMetadata(foreignClassName, key);
+  if(foreignMeta && foreignMeta.dataType) {
+    this.defineProperty(className, key, {
+      type: this.columnDataType(foreignClassName, key),
+      postgresql: {
+        dataType: foreignMeta.dataType
+      }
+    });
+  } else {
+    cb(null, this.columnDataType(foreignClassName, key));
+  }
+}
+
+/*!
  * Build a list of columns for the given model
  * @param {String} model The model name
  * @returns {String}


raymondfeng:
Good catch. A better way is to fix it is probably around https://github.com/strongloop/loopback-datasource-juggler/blob/master/lib/datasource.js#L1703. Please note we derive the type for the foreign key from the primary key. The db specific settings should be honored too. Do you want to create a patch?

aol-nnov:
@raymondfeng okay, I'll look around the lines you noted. Hopefully, tomorrow. I'll open a PR when it's ready.
Btw, is the CLA I've signed for all loopback projects or there is an agreement per project and I should sign another one before committing to loopback-datasource-juggler?

aol-nnov:
closed in favour of strongloop/loopback-datasource-juggler#420

Issue #31, opened 2014-09-29
Does this support the Postgres JSON type? 


gebrits:
I'd like to use Postgres with the 'new' JSON type to allow for flexible/evolving data-structures. Is this supported through this loopback connector?

ezraroi:
What about Array types? How we should define them?

dh376:
+1

geekguy:
Any update on this ?

ssh24:
You can specify a column type to be explicitly postgres `JSON` type of the specific property in the model-definition.

Here is an example code snippet:

**model-definition.json**
```json
{
  "name": "Business",
  "base": "PersistedModel",
  "idInjection": false,
  "options": {
    "validateUpsert": true
  },
  "properties": {
    "id": {
      "type": "number",
      "required": true,
      "id": true
    },
    "name": {
      "type": "string"
    },
    "location": {
      "type": "object",
      "postgresql": {
        "dataType": "json"
      }
    }
  },
  "validations": [],
  "relations": {},
  "acls": [],
  "methods": {}
}
```

**boot-script.js**
```js
'use strict';
var util = require('util');
var _ = require('lodash');

module.exports = function(app) {
    var db = app.datasources.postgresqlDs;
    var Business = app.models.Business;

    db.automigrate(function(err) {
        if (err) throw err;
        console.log('\nAutomigrate completed');

        Business.create([{
            id: 1,
            name: 'Foo Inc.',
            location: {
                'United States': {
                    stores: 10
                },
                'United Kingdom': {
                    stores: 5
                },
                'Brazil': {
                    stores: 2
                }
            }
        }, {
            id: 2,
            name: 'Bar Inc.',
            location: {
                'United States': {
                    stores: 12
                },
                'United Kingdom': {
                    stores: 20
                },
                'Argentina': {
                    stores: 4
                }
            }
        }], function(err, result) {
            if (err) throw err;
            console.log('\nCreated instance: ' + util.inspect(result, {depth: 4}));

            Business.find(function(err, result) {
                if (err) throw err;
                console.log('\nFound instance: ' + util.inspect(result, {depth:4}))
            });
        });
    });
};
```

**output**
```bash
sakibs-mac:postgresql-31 ssh$ node .
Web server listening at: http://0.0.0.0:3000
Browse your REST API at http://0.0.0.0:3000/explorer

Automigrate completed

Created instance: [ { id: 1,
    name: 'Foo Inc.',
    location:
[...]

PR #59, opened 2015-02-09
[wip] native JSON support

aol-nnov:
Hi!

I was i need of JSON data type power which PostgreSQL provides, so I've started working in this direction.
It seems, that there is a demand for that support already: #31, #7, so, @gebrits, @kesavkolla I invite you to collaborate in this area if you still interested.

@raymondfeng what do you think about enabling certain features in the connector (upsert support, json, whatever) based on the database version? Where in the code is the best way to test the DB version and fill a set of supported features? Should it be done automatically or set via config? Also tell me please what do you think about mods in this PR?

### DIFF
--- a/lib/postgresql.js
+++ b/lib/postgresql.js
@@ -477,10 +477,7 @@ PostgreSQL.prototype.toDatabase = function (prop, val) {
     }
   }
   if (val.constructor.name === 'Object') {
-    if (prop.postgresql && prop.postgresql.dataType === 'json') {
-       return JSON.stringify(val);
-    }
-    var operator = Object.keys(val)[0]
+    var operator = Object.keys(val)[0];
     val = val[operator];
     if (operator === 'between') {
       return this.toDatabase(prop, val[0]) + ' AND ' + this.toDatabase(prop, val[1]);
@@ -492,6 +489,9 @@ PostgreSQL.prototype.toDatabase = function (prop, val) {
       return val.join(',');
     }
     return this.toDatabase(prop, val);
+  }
+  if(prop.type.name == 'Json') {
+    return JSON.stringify(val);
   }
   if (prop.type.name === 'Number') {
     if (!val && val !== 0) {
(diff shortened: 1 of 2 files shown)


aol-nnov:
nice articles that helped me a lot:
http://www.postgresql.org/docs/9.3/static/functions-json.html
http://schinckel.net/2014/05/25/querying-json-in-postgres/

lib/postgresql.js
```diff
@@ -477,10 +477,7 @@ PostgreSQL.prototype.toDatabase = function (prop, val) {
     }
   }
   if (val.constructor.name === 'Object') {
-    if (prop.postgresql && prop.postgresql.dataType === 'json') {
```
kesavkolla: PostgreSQL 9.4 introduced new type jsonb.  So add jsonb also in here.

kesavkolla:
Actually my example is bad.  The escapeName doesn't even do anything with select columns.  The escapeName is only called from createTable and columns to be dropped.  So there is no need to touch that code in my opinion.

jmls:
did this stall here ? It would be useful to have the 9.4 json datatypes in place now

aol-nnov:
@jmls sad to say, I've dropped it completely. Moved away from loopback and using bookshelf.js currently.

jmls:
@aol-nnov : That's not good to hear. May I ask why ?

aol-nnov:
@jmls well, it was a long way to go :)
first of all, I had to wrap my head around all the guts loopback have. And the best way was to start something similar from scratch, lol.
secondly, I needed better postgresql support, and, to be frank, I really love bookshelf and knex.
and my app is not only api, but a web frontend too (well, it actually does not cound as an argument and I could build web frontend atop of loopback)

ZepAviator:
Having JSON / JSONb support would be great. It seems that maybe some ideas for code on these properties types could be pulled from the nosql connectors?

Anyone have any branches of working JSON type?

slnode:

slnode:

slnode:

zbarbuto:
FWIW #304 was merged recently

PR #67, opened 2015-03-24
Return count when updating or deleting

superkhau:
- `update/updateAll` returns item count
- `destroyAll` return item count

Connect to strongloop/loopback#1167

### DIFF
--- a/lib/postgresql.js
+++ b/lib/postgresql.js
@@ -119,7 +119,18 @@ PostgreSQL.prototype.executeSQL = function (sql, params, callback) {
       // console.log(err);
       if (log) log(sql, time);
       process.nextTick(done); // Release the pooled client
-      callback(err ? err : null, data ? data.rows : null);
+      var result = null;
+      if (data) {
+        switch(data.command) {
+          case 'DELETE':
+          case 'UPDATE':
+            result = {count: data.rowCount};
+            break;
+          default:
+            result = data.rows;
+        }
+      }
+      callback(err ? err : null, result);
     });
   });
 };


superkhau:
@raymondfeng PTAL

lib/postgresql.js
```diff
@@ -369,11 +375,7 @@ PostgreSQL.prototype.update =
     data = this.mapToDB(model, data);
     var props = this._categorizeProperties(model, data);
 
-    this.query(sql, generateQueryParams(data, props), function (err, result) {
-      if (callback) {
-        callback(err, result);
-      }
-    });
+    this.query(sql, generateQueryParams(data, props), callback);
```
bajtos: This is a breaking change, isn't it? Before your patch, one could call the method without providing any callback.

The correct fix:

```js
callback = callback || function(){};
this.query(sql, generateQueryParams(data, props), callback);
```

lib/postgresql.js
```diff
@@ -119,7 +119,12 @@ PostgreSQL.prototype.executeSQL = function (sql, params, callback) {
       // console.log(err);
       if (log) log(sql, time);
       process.nextTick(done); // Release the pooled client
-      callback(err ? err : null, data ? data.rows : null);
+
+      var affectedCount = data && typeof data.rowCount === 'number' &&
+          !isNaN(data.rowCount) ? data.rowCount : 0;
+
+      callback(err ? err : null, data ? data.rows : null, {count:
+          affectedCount});
```
bajtos: Please pass the full `data` object, it should be the responsibility of individual DAO methods to decide what properties of the pg response should be passed on to the caller and how they should be transformed.

I also think you should return `undefined` instead of `0` if `rowCount` is not a valid number. That way you will signal that the database did not correctly report number of affected records. Your current implementation silently hides this problem, it will be extremely difficult to troubleshoot consequences of this.

raymondfeng: +1 to callback with the full `data` object and remove the 3rd arg.

We should use `undefined` or `-1` if the count is unknown.

superkhau:
@raymondfeng PTAL again, this is the simplest way to preserve BC while massaging the response/second arg for `updateAll/destroyAll`.

raymondfeng:
LGTM

superkhau:
@slnode test please

superkhau:
@slnode test please

superkhau:
@slnode test please

PR #69, opened 2015-03-25
Changes in PostgreSQL datatypes

yellowiscool:


slnode:

bajtos:
@slnode ok to test

bajtos:
@yellowiscool thank you for the pull request. I am not familiar enough with this project, I'll let @raymondfeng to review your patch.

However, I would like to point out a possible backwards compatibility issue. IIUC, this change is forcing all users of the PostgreSQL connector to modify their database schema to use these new types, at least when they run auto-update/auto-migrate script. 

Perhaps we can add a flag (connector/datasource setting) to allow the users to decide whether they want the old or the new mapping?

kadim:
Why not JSONB? It can be indexed!

TinOo512:
@raymondfeng can we have update on this ? If I read the postgresql documentation (http://www.postgresql.org/docs/9.1/static/datatype-character.html) there is no benefits to use `character varying` with a specified length.

slnode:

tobernguyen:
Why this pull request take so long? Please review and accept this so this problem will be fixed.
https://github.com/strongloop/loopback-component-passport/issues/84

slnode:

superkhau:
Can you add some tests to verify your changes and prevent regressions in the future?

archenroot:
Not sure if relevant to put here, but as it refers to new data types, better do, we have also hstore now, similar to redis key-value fundamentally:
https://github.com/strongloop/loopback-connector-postgresql/issues/232

slnode:

slnode:

slnode:

PR #79, opened 2015-05-17
LIKE -> ILIKE

rus0000:
"ILIKE" case-insensitive, and more suitable for autocomplete comboboxes
http://www.postgresql.org/docs/9.4/static/functions-matching.html

### DIFF
--- a/lib/postgresql.js
+++ b/lib/postgresql.js
@@ -353,12 +353,12 @@ PostgreSQL.prototype.buildExpression =
     switch(operator) {
       case 'like':
         return new ParameterizedSQL({
-          sql: columnName + " LIKE ? ESCAPE '\\'",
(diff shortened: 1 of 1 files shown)


slnode:

bajtos:
Hi @rus0000, thank you for the pull request. In my opinion, this is a breaking change that also introduces inconsistency with the behaviour of other connectors (e.g. `memory` connector). I think it would be better to introduce a new operator instead. We already have `like` and `nlike`, so it should be relatively straightforward to implement `ilike` too.

Having said that, I'll let @raymondfeng review this patch as he is more familiar with the code base.

bajtos:
@slnode ok to test

rus0000:
https://github.com/strongloop/loopback-connector-postgresql/pull/93

slnode:

Issue #94, opened 2015-07-05
isActual always returning false, even after autoupdate


cooperjbrandon:
`isActual` method always returns false. To expand, i extended the base User model, adding `firstname`. I traced the problem down to the dataTypeChanged function, which always returns true for every string datatype (`username`, `password`, `email`, etc...).

This is due to this line: 
https://github.com/strongloop/loopback-connector-postgresql/blob/master/lib/migration.js#L221

Look at the `oldSettings` object, which comes from the `actualFields` parameter (which comes from these db query: https://github.com/strongloop/loopback-connector-postgresql/blob/master/lib/migration.js#L23-26). The `type` property of `oldSettings` for any string is `character varying`, but the `self.columnDataType` evalutes to `VARCHAR (1024)`, from this line: 
https://github.com/strongloop/loopback-connector-postgresql/blob/master/lib/migration.js#L466.

Essentially, `CHARACTER VARYING` will never equal `VARCHAR (1024)`, even after running autoupdate. Is there a work around for this?

I also notice there is the `postgresqlDataTypeToJSONType` function, but it isn't being used anywhere.

cooperjbrandon:
resolved - see this commit: e04ab44f30d2f398c6ecec86fd51c8827918801a

shSina:
double is not data type and cause this function to always return false , when we use double precision as data type.

PR #95, opened 2015-07-07
include the length of the column if its a char type. change character…

cooperjbrandon:
… varying to varchar to match columnDataType method

`columnDataType` method for type of `string` returns `VARCHAR(<colLength>)`.  This never matches the `oldSettings.type` result in `datatypeChanged` function for two reasons:
1) `oldSettings.type` doesn't return the length of the column.
2) `oldSettings.type` could possibly be `CHARACTER VARYING`, which clearly doesn't === `VARCHAR`.

This commit fixes the problem and addresses this issue:
https://github.com/strongloop/loopback-connector-postgresql/issues/94

### DIFF
--- a/lib/migration.js
+++ b/lib/migration.js
@@ -14,14 +14,14 @@ function mixinMigration(PostgreSQL) {
       }
       if (!err) {
         data.forEach(function(field) {
-          field.type = mapPostgreSQLDatatypes(field.type);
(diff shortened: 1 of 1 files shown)


cooperjbrandon:
?

raymondfeng:
ok to test

PR #96, opened 2015-07-21
Add support for RegExp operator

superkhau:
Connect to strongloop/loopback-datasource-juggler#552

### DIFF
--- a/lib/postgresql.js
+++ b/lib/postgresql.js
@@ -354,34 +354,34 @@ PostgreSQL.prototype.applyPagination = function(model, stmt, filter) {
   return stmt.merge(limitClause);
 };
 
-PostgreSQL.prototype.buildExpression =
-  function(columnName, operator, columnValue, propertyValue) {
-    if (propertyValue instanceof RegExp) {
-      columnValue = "'" + propertyValue.source + "'";
-      if (propertyValue.ignoreCase) {
-        return new ParameterizedSQL(columnName + ' ~* ?', [columnValue]);
-      } else {
-        return new ParameterizedSQL(columnName + ' ~ ?', [columnValue]);
-      }
+PostgreSQL.prototype.buildExpression = function(columnName, operator,
+    columnValue, propertyValue) {
+  if (propertyValue instanceof RegExp) {
+    columnValue = "'" + propertyValue.source + "'";
+    if (propertyValue.ignoreCase) {
+      return new ParameterizedSQL(columnName + ' ~* ?', [columnValue]);
+    } else {
+      return new ParameterizedSQL(columnName + ' ~ ?', [columnValue]);
     }
-    switch(operator) {
-      case 'like':
-        return new ParameterizedSQL({
-          sql: columnName + " LIKE ? ESCAPE '\\'",
-          params: [columnValue]
-        });
-      case 'nlike':
-        return new ParameterizedSQL({
-          sql: columnName + " NOT LIKE ? ESCAPE '\\'",
-          params: [columnValue]
-        });
-      default:
-        // Invoke the base implementation of `buildExpression`
-        var exp = this.invokeSuper('buildExpression',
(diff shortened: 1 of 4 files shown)


rmg:
I haven't looked, but there is a good chance the build failures are the result of the tests running at the same time as the master branch tests - these tests are not safe to run in parallel because they are using a shared database.

lib/postgresql.js
```diff
@@ -354,34 +354,41 @@ PostgreSQL.prototype.applyPagination = function(model, stmt, filter) {
...
-          params: [columnValue]
-        });
-      default:
-        // Invoke the base implementation of `buildExpression`
-        var exp = this.invokeSuper('buildExpression',
-          columnName, operator, columnValue, propertyValue);
-        return exp;
+PostgreSQL.prototype.buildExpression = function(columnName, operator,
+    columnValue, propertyValue) {
+  if (operator === 'regexp' && typeof columnValue === 'string') {
```
raymondfeng: Is there any reason that you don't want make this block part of the switch/case?

lib/postgresql.js
```diff
@@ -354,34 +354,41 @@ PostgreSQL.prototype.applyPagination = function(model, stmt, filter) {
...
+    if (columnValue.charAt(0) === '/')
```
raymondfeng: What is this test for?

superkhau: Yes, because the columnValue returned by the parser gives me `/^a/` when a regex object is provided.

lib/postgresql.js
superkhau: No, I just saw the existing RegExp block and put mine ahead of it.

raymondfeng:
@superkhau Please clean up the code based on my comments.

superkhau:
@raymondfeng Please review again.

lib/postgresql.js
```diff
@@ -6,6 +6,7 @@ var SqlConnector = require('loopback-connector').SqlConnector;
 var ParameterizedSQL = SqlConnector.ParameterizedSQL;
 var util = require('util');
 var debug = require('debug')('loopback:connector:postgresql');
+var utils = require('loopback-datasource-juggler/lib/utils');
```
raymondfeng: This is invalid. `loopback-datasource-juggler` is dev dep only.

lib/postgresql.js
```diff
@@ -354,34 +355,48 @@ PostgreSQL.prototype.applyPagination = function(model, stmt, filter) {
...
-      columnValue = "'" + propertyValue.source + "'";
-      if (propertyValue.ignoreCase) {
-        return new ParameterizedSQL(columnName + ' ~* ?', [columnValue]);
-      } else {
-        return new ParameterizedSQL(columnName + ' ~ ?', [columnValue]);
-      }
+PostgreSQL.prototype.buildExpression = function(columnName, operator,
+    columnValue, propertyValue) {
+  if (operator === 'regexp') {
+    var regex = utils.getRegExpExpression(utils.toRegExp(columnValue));
```
raymondfeng: The regexp value is already an instance of `RegExp`. Why further processing here?

lib/postgresql.js
superkhau: Should I dupiicate the functions then? I assume including juggler is bad.

lib/postgresql.js
superkhau: This is a copy/paste mistake. Removed.

raymondfeng: RegExp should have the methods for what you need.

superkhau:
@raymondfeng LGTY?

raymondfeng:
LGTM

superkhau:
@raymondfeng Updated this PR to match MySQL. Gonna merge without review as its the same, but prints and test console.warn messages now.

PR #97, opened 2015-07-29
Add regexp operator tests

superkhau:
Add tests to ensure regexp operator works properly when flags are provided.

### DIFF
--- a/test/postgresql.test.js
+++ b/test/postgresql.test.js
@@ -206,6 +206,15 @@ describe('postgresql connector', function () {
           console.warn.restore();
         });
 
+        it('should work', function(done) {
+          Post.find({where: {content: {regexp: '^a/i'}}}, function(err, posts) {
+            should.not.exist(err);
+            posts.length.should.equal(1);
+            posts[0].content.should.equal('AAA');
+            done();
+          });
+        });
+
         it('should print a warning when the global flag is set',
             function(done) {
           Post.find({where: {content: {regexp: '^a/g'}}}, function(err, posts) {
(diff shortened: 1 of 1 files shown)


superkhau:
@raymondfeng Merging without review, hardening with more tests.

test/postgresql.test.js
```diff
@@ -206,6 +206,15 @@ describe('postgresql connector', function () {
           console.warn.restore();
         });
 
+        it('should work', function(done) {
+          Post.find({where: {content: {regexp: '^a/i'}}}, function(err, posts) {
```
raymondfeng: Is the leading `/` optional? JS uses `/^a/i` as the literal.

superkhau: For regex strings it is. My toRegExp normalization will convert it to a RegExp object properly (which in turn converts '^a/i' to /a/i underneath the hood.

superkhau: Also its not really just the leading slash, its more "you don't need to provide regex delimiters for regex strings" (meaning the leading and trailing forward slashes).

raymondfeng: ok, I wonder if we want to be strict, i.e., the string should be:

1. /pattern/
2. /pattern/i
3. pattern

I'm fine to allow `pattern/i` if it doesn't cause confusions.

superkhau: Hm...good question. `pattern/i` is actually the syntax MySQL uses. I believe I coded defensively, so maybe just don't announce it in the docs, but let it slide?

PR #98, opened 2015-07-29
Fix regexp operator test

superkhau:
Forgot to convert regex literals to regex objects in tests.


superkhau:
@raymondfeng Merging without review. Forgot to convert some regex literals to regex objects.

PR #99, opened 2015-07-29
Clean up regexp operator tests

superkhau:
Wrap long lines in regexp operator tests.


superkhau:
@raymondfeng Merging without review. Wrapping long lines in tests.

PR #119, opened 2015-11-18
add CI DB server

cgole:


test/init.js
```diff
@@ -2,6 +2,16 @@ var DataSource = require('loopback-datasource-juggler').DataSource;
 
 var config = require('rc')('loopback', {test: {postgresql: {}}}).test.postgresql;
 
+if (process.env.CI) {
+  config = {
+    host: process.env.POSTGRESQL_HOST || config.host || 'localhost',
+    port: process.env.POSTGRESQL_PORT || config.port || 5432,
+    database: 'test',
+    username: process.env.POSTGRESQL_USERNAME || config.username,
```
rmg: Should be `POSTGRESQL_USER`, not `POSTGRESQL_USERNAME`.

rmg:
Tests are failing, possibly because it isn't using the right username.

cgole:
My changes shouldn't affect our strongloop CI at all.  Since the env variable POSTGRESQL_USERNAME is not set it should just use the local db or the one in config.

rmg:
StrongLoop CI sets `CI`, so this does change how this connector's tests run there. And in our CI, it uses `POSTGRESQL_USER` to give the username.

cgole:
test please

cgole:
test please

rmg:
@cgole the tests are failing because the tests expect the database `test` to already exist instead of creating it first.

cgole:
test please

cgole:
test please

rmg:
Not sure the `process.env.CI` check is required, but LGTM.

rmg:
@cgole you should rebase/squash your branches before merging them so that the commits tell a clear story. In this case, it probably should have been a single commit.

Issue #131, opened 2016-03-16
loopback will not force a `postgresql` dataType for embedded relation


sidewaiise:
When using `embedOne` relation type, we cannot specify to use a postgresql `dataType`: "TEXT" and have autoUpdate update the column. The column defaults to varchar(1024).

Here is the sample JSON:

```
    "agentSignature": {
      "type": "embedsOne",
      "model": "Signature",
      "property": "agentSignature",
      "options" : {
        "validate": true,
        "forceId": false
      },
      "postgresql": {
        "dataType": "TEXT",
        "nullable": "YES"
      }
    }
```

and here is an example of the generated database column:

![screen shot 2016-03-16 at 1 18 11 pm](https://cloud.githubusercontent.com/assets/5085804/13800588/837f5456-eb7a-11e5-8a7e-7e688ebe54bb.png)

Essentially, we need the column to store long TEXT strings in JSON format, the JSON being a row of the Signature model.

This seems to be a problem with the way loopback-connector-postgresql handles embedded relations - it doesn't seem to take tabled data structures into account when using nested return values. This may also be a problem with loopback-datasource-juggler.

Any insight into the issue or fixes to the repo would be much appreciated.

Cheers.

madianas:
What about embedsMany? I am having the same issue. It's not picking up "TEXT" datatype

>     "tierprograms": {
      "type": "embedsMany",
      "model": "Tier",
      "property": "tiers",
      "options": {
        "validate": true,
        "forceId": true,
        "persistent": true
      },
      "postgresql": {
        "dataType": "text",
        "nullable": "YES"
      }

lius:
I'm using embedsMany and embedsOne without problems, with json and jsonb data types on database (I did not use the schema generation feature). Also, I'm not specifying a data type specific for postgres per column, just using LB types array and object for model properties. In that case, the values are serialised into strings, working perfectly with json/jsonb types, that works receiving strings as an input.

0candy:
Closing this issue as fix has been merged.

PR #132, opened 2016-03-16
- Adds support for long JSON strings when using `embedOne`

jimmylimm:
Without this change, we are unable to use postgresql as a datasource for some model relations that use the `embedsOne`. This is because the previous default value for detected JSON data types is `VARCHAR(1024)`. Now by default, the column data type will be `text`, allowing for long JSON strings to be imported into the database.

References issue:  https://github.com/strongloop/loopback-connector-postgresql/issues/131

### DIFF
--- a/lib/migration.js
+++ b/lib/migration.js
@@ -526,10 +526,10 @@ function mixinMigration(PostgreSQL) {
     switch (prop.type.name) {
       default:
       case 'String':
-      case 'Text':
-        return 'VARCHAR' + (colLength ? '(' + colLength + ')' : '(1024)');
       case 'JSON':
         return 'TEXT';
+      case 'Text':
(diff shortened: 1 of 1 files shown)


slnode:

lib/migration.js
```diff
@@ -527,7 +527,7 @@ function mixinMigration(PostgreSQL) {
       default:
       case 'String':
       case 'JSON':
-        return 'VARCHAR' + (colLength ? '(' + colLength + ')' : '(1024)');
+        return 'TEXT';
       case 'Text':
         return 'VARCHAR' + (colLength ? '(' + colLength + ')' : '(1024)');
```
raymondfeng: This should be TEXT too, right?

jimmylimm: Yes agree, will update now.

jimmylimm: All updated.

superkhau:
LGTM

PR #140, opened 2016-04-30
Lazy connect when booting from swagger

jannyHou:
Connect to strongloop/loopback#2242

When booting app from swagger generator, datasources are setup with ds.lazyConnect = true to avoid connection.


jannyHou:
@raymondfeng Please review:)

raymondfeng:
LGTM

Issue #123, opened 2016-01-12
When new required fields with a default value are added autoupdate shouldn't fail


alFReD-NSH:
Steps to produce:

1 - Create a model and run autoupdate/automigrate
2 - Add a new required property with a default value to the model for example:

```
    "newProperty": {
      "type": "Number",
      "required": true,
      "default": 0
    }
```

3 - Run autoupdate again

What's expected: New column should be added with the default value.

What actually happens: Autoupdate fails here and throws this:

```
error: column "newProperty" contains null values
```

It happens because it tries to add column with `NOT NULL` constraint, but this is only possible, if we give a default value, but `loopback-connector-postgresql` ignores the given default value in the model configuration.

I would like to contribute this myself, but I'm not sure if I'm getting the test configuration to work. I'm getting `error: password authentication failed for user "user"` for the first test, but the credentials in the configuration are correct and also a lot of other tests are failing.

Elindorath:
@raymondfeng @superkhau @bajtos : I would like to address this issue, but I'm unable to run the test right now. I saw on the Readme that I need to contact you for instructions so I do.

raymondfeng:
@Elindorath You have two ways to configure your postgres connection.

1. Set up env vars (https://github.com/strongloop/loopback-connector-postgresql/blob/master/test/init.js#L10)
```
export CI=true
export TEST_POSTGRESQL_HOST=localhost
export TEST_POSTGRESQL_PORT=5432
export TEST_POSTGRESQL_USER=myUser
export TEST_POSTGRESQL_PASSWORD=myPassword
```

Or drop `.loopbackrc` file into your home dir:
```
{
    "test": {
        "postgresql": {
            "host": "localhost",
            "port": 5432,
            "database": "test",
            "username": "test",
            "password": "testpass"
        }
    }
}
```

Please also use https://github.com/strongloop/loopback-connector-postgresql/blob/master/test/tables.sql to populate your test db.

Elindorath:
Thanks @raymondfeng. I'll try to give you an update this week-end.

zihehuang:
Any update on this issue? I am running into the same issue.

kutsyk:
Is this fixed?

bvanleeuwen1995:
To me this seems to be still a bug, either defaultFn or default is not honerred during migrations and it will fail.

ptmccarthy:
I'm still running into this issue today

bartshappee:
[...]

PR #153, opened 2016-07-27
Fixes #123: Sets default value for added columns when doing an autoupdate.

MichaelMarner:
If a model property has a default value, set that as the database column's default value when adding the column in an auto update.

Patch modifies the SQL generated for adding a new column. If the loopback model has a default value, it uses this in the ADD COLUMN statement.

This ensures that existing database rows will have a value for the new column, and prevents the autoupdate failing due to null values.

Fixes #123

### DIFF
--- a/lib/migration.js
+++ b/lib/migration.js
@@ -218,7 +218,13 @@ function mixinMigration(PostgreSQL) {
       var found = self.searchForPropertyInActual(
         model, self.column(model, propName), actualFields);
       if (!found && self.propertyHasNotBeenDeleted(model, propName)) {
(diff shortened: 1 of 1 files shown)


slnode:

slnode:

superkhau:
Can you add some tests to verify your changes and prevent regressions in the future?

slnode:

superkhau:
Thanks for the contribution! ;) I just landed https://github.com/strongloop/loopback-connector-postgresql/pull/167 which appears be a duplicate of this PR. Closing as duplicate -- feel free to ping me if you feel otherwise and I'll reopen.

PR #158, opened 2016-08-10
test: allow testing against local database

rmg:
Same basic idea as was done in strongloop/loopback-connector-mysql#184 and strongloop/loopback-connector-mysql#186 except cleaner because Postgres isn't MySQL.


superkhau:
LGTM

PR #166, opened 2016-09-07
Use juggler@3 for running tests

0candy:
Connect to https://github.com/strongloop-internal/scrum-loopback/issues/1040

### DIFF
--- a/lib/postgresql.js
+++ b/lib/postgresql.js
@@ -14,6 +14,7 @@ var SqlConnector = require('loopback-connector').SqlConnector;
 var ParameterizedSQL = SqlConnector.ParameterizedSQL;
 var util = require('util');
 var debug = require('debug')('loopback:connector:postgresql');
+var Promise = require('bluebird');
 
 /**
  *
@@ -79,6 +80,7 @@ function PostgreSQL(postgresql, settings) {
   } else {
     this.clientConfig = settings;
   }
+  this.clientConfig.Promise = Promise;
   this.pg = new postgresql.Pool(this.clientConfig);
   this.settings = settings;
   if (settings.debug) {
--- a/package.json
+++ b/package.json
@@ -18,16 +18,16 @@
   },
   "dependencies": {
     "async": "^0.9.0",
+    "bluebird": "^3.4.6",
     "debug": "^2.1.1",
     "loopback-connector": "^2.1.0",
     "pg": "^6.0.0",
     "strong-globalize": "^2.6.2"
   },
   "devDependencies": {
-    "bluebird": "^2.9.12",
     "eslint": "^2.13.1",
     "eslint-config-loopback": "^4.0.0",
-    "loopback-datasource-juggler": "^2.28.0",
+    "loopback-datasource-juggler": "^3.0.0-alpha.8",
     "mocha": "^2.1.0",
     "rc": "^1.0.0",
     "should": "^8.0.2",


rmg:
I'm pretty sure I've tracked down the problem. I'll be deploying a fix soon (within a few hours, hopefully).

rmg:
@slnode test please

rmg:
cis-jenkins builds are now working - and failing on node 0.10 due to `Promise` not being defined.

rmg:
@slnode test please

superkhau:
>cis-jenkins builds are now working - and failing on node 0.10 due to Promise not being defined.

@0candy We can take care of this one, ping me on slack if you run into any issues.

0candy:
@slnode test please

0candy:
@rmg I think postgres is down on the windows machine.

rmg:
@slnode test please

rmg:
Had to fix a couple issues with the Windows environments. Will hopefully work on the next build.

@slnode test please

lib/postgresql.js
```diff
@@ -15,6 +15,11 @@ var ParameterizedSQL = SqlConnector.ParameterizedSQL;
 var util = require('util');
 var debug = require('debug')('loopback:connector:postgresql');
 
+// Add support for node 0.10.x
+if (typeof Promise == 'undefined') {
+  global.Promise = require('promise-polyfill');
```
bajtos: We usually use `bluebird` instead of `promise-polyfill`. I am proposing to use `bluebird` here too, for the sake of consistency.

Thoughts?

bajtos: This is actually weird.

https://github.com/strongloop/loopback-datasource-juggler/blob/master/3.0-RELEASE-NOTES.md#always-use-bluebird-as-promise-library
> In version 3.0, we always use bluebird as our promise library instead of `global.Promise`. We consider Bluebird API a part of LoopBack API from now on, you are welcome to use any Bluebird-specific methods in your applications.

When using juggler@3, it should not be needed to patch `global.Promise`. Unless there is a rouge test in juggler that uses `global.Promise` instead of `require('bluebird')`, in which case we need to make the fix in juggler.

@superkhau @0candy could you PTAL?

0candy: > cis-jenkins builds are now working - and failing on node 0.10 due to Promise not being defined.

I was looking into it and `pg` has `pg-pool` which states this:
https://www.npmjs.com/package/pg-pool#bring-your-own-promise

superkhau: We should be using bluebird for v0.10, Promise is not defined because native promises (global.Promise) are not supported in v0.10.

0candy: > We usually use bluebird instead of promise-polyfill. I am proposing to use bluebird here too, for the sake of consistency.

I tried changing it to `promise` earlier but CI still failed.

superkhau:
test please

lib/postgresql.js
```diff
@@ -15,6 +15,9 @@ var ParameterizedSQL = SqlConnector.ParameterizedSQL;
 var util = require('util');
 var debug = require('debug')('loopback:connector:postgresql');
 
+// Add support for node 0.10.x
```
superkhau: remove

package.json
```diff
@@ -18,16 +18,16 @@
...
     "loopback-connector": "^2.1.0",
     "pg": "^6.0.0",
     "strong-globalize": "^2.6.2"
   },
   "devDependencies": {
-    "bluebird": "^2.9.12",
     "eslint": "^2.13.1",
     "eslint-config-loopback": "^4.0.0",
-    "loopback-datasource-juggler": "^2.28.0",
+    "loopback-datasource-juggler": "^3.0.0-alpha.7",
```
superkhau: alpha.8

superkhau:
I think on https://github.com/strongloop/loopback-connector-postgresql/blob/master/lib/postgresql.js#L77, you can add `Promise: require('bluebird')` to the clientConfigs for the fix.

bajtos:
Nice, I like the final version 👍

PR #176, opened 2016-10-07
Remove unused prefix from test env vars

superkhau:
* TEST_ prefix is not used by CI anymore and therefore serve no purpose

* Part of overall goal to standardize env var injection to
  MODULE_VARNAME (ie. MONGODB_HOST for example) convention

Connect to https://github.com/strongloop-internal/scrum-loopback/issues/1096

cc @bajtos @rmg @strongloop/fa-db-connectors

### DIFF
--- a/test/init.js
+++ b/test/init.js
@@ -9,18 +9,22 @@ var DataSource = require('loopback-datasource-juggler').DataSource;
 var config = require('rc')('loopback', {test: {postgresql: {}}}).test.postgresql;
 
 if (process.env.CI) {
-  process.env.PGHOST = process.env.TEST_POSTGRESQL_HOST ||
-    process.env.POSTGRESQL_HOST || process.env.PGHOST || 'localhost';
-  process.env.PGPORT = process.env.TEST_POSTGRESQL_PORT ||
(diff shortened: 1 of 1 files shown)


test/init.js
```diff
@@ -9,18 +9,22 @@ var DataSource = require('loopback-datasource-juggler').DataSource;
...
+    process.env.PGUSER ||
+    'test';
+  process.env.PGPASSWORD = process.env.POSTGRESQL_PASSWORD ||
+    process.env.PGPASSWORD ||
+    '';
   config = {
     host: process.env.PGHOST,
     port: process.env.PGPORT,
-    database: process.env.TEST_POSTGRESQL_DATABASE || 'emptytest',
+    database: process.env.POSTGRESQL_DATABASE || 'emptytest',
```
bajtos: should we support `PGDATABASE` too, for consistency with other `PG*` env vars?

superkhau: Outside the scope of this PR, but IMO yes we should add it (I'll send another PR after landing this).

superkhau: https://github.com/strongloop/loopback-connector-postgresql/pull/178

PR #178, opened 2016-10-07
Accept PGDATABASE env var in test/init.js

superkhau:
See https://github.com/strongloop/loopback-connector-postgresql/pull/176#discussion_r82351420

@bajtos PTAL

cc @strongloop/fa-db-connectors


bajtos:
LGTM. I am happy to delegate code reviews to @strongloop/fa-db-connectors for the future.

PR #187, opened 2016-11-01
Update README with correct doc links, etc

Amir-61:
#* Update README with correct doc links, etc

/to: @superkhau @crandmck

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,8 +1,8 @@
-## loopback-connector-postgresql
+# loopback-connector-postgresql
 
-The PostgreSQL Connector module for for [loopback-datasource-juggler](http://docs.strongloop.com/loopback-datasource-juggler/).
+The official PostgreSQL connector for the LoopBack framework.
 
-Please see the [official documentation](http://docs.strongloop.com/display/LB/PostgreSQL+connector).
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1,8 +1,8 @@
-## loopback-connector-postgresql
+# loopback-connector-postgresql
 
-The PostgreSQL Connector module for for [loopback-datasource-juggler](http://docs.strongloop.com/loopback-datasource-juggler/).
+The PostgreSQL Connector module for [loopback-datasource-juggler](http://docs.strongloop.com/loopback-datasource-juggler/).
```
Amir-61: **Please Note:**

I could not find the equivalent of http://docs.strongloop.com/loopback-datasource-juggler/ in http://loopback.io/doc/en/lb2/

Could you help?

superkhau: I think its supposed to connect to the Juggler repo like https://github.com/strongloop/loopback-connector-mysql/blame/master/README.md#L3. However, I think we should just change the entire sentence:

"The official PostgreSQL connector for the LoopBack framework" would suffice. See https://github.com/strongloop/loopback-connector-kv-redis#loopback-connector-kv-redis

Issue #191, opened 2016-11-29
geo near filter is not supported


gregdingle:
Despite postgres having strong support for geo functions through the postgis extension, the `near` filter is not translated into SQL. This issue could also be related to #126.

The juggler then will calculate distance post-query, which is sub-optimal. See https://github.com/strongloop/loopback-datasource-juggler/blob/master/lib/dao.js#L1794 . 

The mongodb connector has support: https://github.com/strongloop/loopback-connector-mongodb/blob/a44aec12572111d86d622c4083e9dbed78e739b0/lib/mongodb.js#L759

It looks like we just need to write a `buildNearFilter` function in `postgresql.js` that constructs the SQL functions: 
* postgis distance: http://postgis.net/docs/geometry_distance_knn.html
* postgis distance within: http://postgis.net/docs/ST_DWithin.html

FreakTheMighty:
I think I've hit this issue. I didn't realize that distance in the `near` query is calculated in the application layer. It appears to be causing large CPU spikes.

Perhaps while this pull request is pending https://github.com/strongloop/loopback-connector-postgresql/pull/192, we can add a note that while `GeoPoint` is supported, the `near` query has limitations.

gregdingle:
@FreakTheMighty : an intern on my team implemented a quick implementation of #192 . See https://github.com/gregdingle/loopback-connector-postgresql/compare/near-filter...Bubblyworld:near-filter . However, he did not implement test cases nor did I review the code yet. And he has gone back to school 🙃

FreakTheMighty:
@gregdingle thank you. At very least this will be a very good place to start.

PR #192, opened 2016-11-30
Add near filter #191

gregdingle:
**This is an RFC pull request: request for comment.** It addresses #191 . I read the connector code and identified all the places where changes were needed. Please read the **TODOs** and tell me if my reasoning is correct. Then I can (or someone else) can continue with the implementation.

### DIFF
--- a/lib/postgresql.js
+++ b/lib/postgresql.js
@@ -86,6 +86,13 @@ function PostgreSQL(postgresql, settings) {
   if (settings.debug) {
     debug('Settings %j', settings);
   }
+
+  //TODO (gdingle): create setting for postgis or check for it dynamically
+  //  if (this.settings.enableGeoIndexing === true) {
+  //   PostgreSQL.prototype.buildNearFilter = buildNearFilter;
+  // } else {
+  //   PostgreSQL.prototype.buildNearFilter = undefined;
+  // }
 }
 
 // Inherit from loopback-datasource-juggler BaseSQL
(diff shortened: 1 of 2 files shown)


slnode:

slnode:

slnode:

slnode:

gregdingle:
@superkhau ?

superkhau:
@gregdingle Took a quick look, yes you're on the right track. The only thing I question is when enableGeoIndexing is set to undefined -- we need a way to tell the user it's not supported. So consider setting it to a function that returns an error instead.

```
Postgresql.prototype.buildNearFilter = function(..., cb) {
  return cb(new Error('buildNearFilter is not supported unless you set enableGeoIndexing: true in ...);
};
```

gregdingle:
@superkhau To be clear, would this pull request be accepted if I implement it and the same test cases that are in the mongdb connector?

gregdingle:
And what should the correct policy be for postgres instances that do **not** have the postgis extension installed and turned on--the default? Should we try to auto-detect at runtime, put a warning in the documentation, or throw an error?

superkhau:
>To be clear, would this pull request be accepted if I implement it and the same test cases that are in the mongdb connector?

Your new PR doesn't have to be the same tests, in fact they should be developed independently according to the Postgres implementation (meaning they could cover less ground and we can improve incrementally over time).

That said, some consistency would be nice.

>And what should the correct policy be for postgres instances that do not have the postgis extension installed and turned on--the default? Should we try to auto-detect at runtime, put a warning in the documentation, or throw an error?

We should do all of that, except throwing an error -- return it via callback so the caller can handle during runtime. That of course requires a way to detect it at runtime (and possibly boot time too). The default is debatable, default to on and return error if the backend doesn't support is safe IMO.

@bajtos @raymondfeng @ritch Thoughts?

bajtos:
Hello. First of all, thank you @gregdingle for taking a stab at this missing but valuable feature!

> And what should the correct policy be for postgres instances that do not have the postgis extension installed and turned on--the default? Should we try to auto-detect at runtime, put a warning in the documentation, or throw an error?

What happens when a postgis query is sent to such instance? Will the error message returned by PostgreSQL provide enough information for the user to figure out what's wrong and how to fix it? If yes, then I think that's all we need for the first iteration (this pull request).

Auto-detection at runtime would be nice too. I think the preferable solution is to detect postgis availability at the time when the datasource is created and/or the model is attached to this datasource, depending on where the LoopBack feature flag is defined. The idea is that if the database server does not provide all features needed by the app, then the user should be notified immediately at startup time, so that they don't have to wait until eventually a geo query fails.

Adding a warning to documentation is good too, but since most people don't read documentation, documentation changes only are not sufficient.

gregdingle:
@bajtos Thanks for the feedback. I haven't forgotten about this feature though I have got swept up in other things. I'll return to it when it is urgent that my app get proper postgis support. In the meantime, anybody is welcome to continue the implementation I started here.

gregdingle:
An intern on my team implemented a quick implementation of this RFC. See https://github.com/gregdingle/loopback-connector-postgresql/compare/near-filter...Bubblyworld:near-filter . He did not implement test cases nor did I review the code yet. And he has gone back to school 🙃

ssh24:
@gregdingle Any update on this PR? Can we proceed with this?

gregdingle:
Sorry, I got involved in other things. Go ahead without me.

PR #203, opened 2017-01-19
Use unique param for affectedRows

loay:
connect to https://github.com/strongloop/loopback-connector-postgresql/issues/204

### DIFF
--- a/lib/postgresql.js
+++ b/lib/postgresql.js
@@ -153,7 +153,7 @@ PostgreSQL.prototype.executeSQL = function(sql, params, options, callback) {
         switch (data.command) {
           case 'DELETE':
           case 'UPDATE':
-            result = {count: data.rowCount};
+            result = {affectedRows: data.rowCount, count: data.rowCount};
             break;
           default:
             result = data.rows;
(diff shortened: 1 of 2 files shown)


loay:
@bajtos PTAL. Thanks

lib/postgresql.js
```diff
@@ -153,7 +153,7 @@ PostgreSQL.prototype.executeSQL = function(sql, params, options, callback) {
         switch (data.command) {
           case 'DELETE':
           case 'UPDATE':
-            result = {count: data.rowCount};
+            result = {affectedRows: data.rowCount};
```
bajtos: Please preserve the `count` property for backwards compatibility.

```js
result = {affectedRows: data.rowCount, count: data.rowCount};
```

bajtos:
Let's leave the tests out of scope of this patch, since they have to be implemented in a different repository anyways. However, I think it makes sense to add a postgresql-specific test to verify that the `count` property is preserved.

lib/postgresql.js
loay: Done and unit test added

test/postgresql.test.js
```diff
@@ -90,6 +90,21 @@ describe('postgresql connector', function() {
       });
   });
 
+  it('should reserve property `count` after query execution', function(done) {
```
bajtos: `preserve`

test/postgresql.test.js
```diff
@@ -90,6 +90,21 @@ describe('postgresql connector', function() {
...
+    Post.create(
+      {title: 'T10', content: 'C10'},
+      function(err, p) {
+        should.not.exists(err);
```
bajtos: `if (err) return done(err);`

PR #213, opened 2017-02-16
Fix database name if on CI

ssh24:
Select the right database to run discovery/migration test if ran on CI versus local


ssh24:
Unnecessary change. Closing.

Issue #129, opened 2016-03-08
Transaction timeout not preventing DB calls


doorty:
I'm following the loopback transaction doc: https://docs.strongloop.com/display/public/LB/Using+database+transactions

1) I set a short timeout for testing purposes. 
2) After the timeout, the DB calls continue to complete successfully, no errors. I can confirm that create/update were stored in DB.
3) I then call transaction.commit, which returns an error of an empty object, {}
4) But the DB still has all the records as if the timeout didn't happen.

PR #230, opened 2017-03-20
Fix operations on ended transactions

zbarbuto:
### Description

*Issue*

This fixes a pretty critical issue where operations can occur on non-transactional connections when the user expects them to land in a transaction. Tests added which fail on master but pass here.

*Example use case*:

- Begin a transaction.
- Perform some creates on the transaction
- An external event (e.g. timeout) triggers a rollback
- Additional creates are allowed to continue despite the transaction no longer being active
- These creates still hit the database when the programmer expects them not to as the transaction has rolled back (finished)

*Solution*:

I've added a uuid to each transaction which exists on the transaction itself and is tagged onto the current connection. Whenever the transaction is committed or rolled back ('finished') the id is removed from the connection but remains on the transaction. Any attempts to execute sql on the transaction checks the transaction id against the specified connection and will fail if the ids do not match.

#### Related issues

- #129

### Checklist

### DIFF
--- a/lib/postgresql.js
+++ b/lib/postgresql.js
@@ -197,12 +197,12 @@ PostgreSQL.prototype.executeSQL = function(sql, params, options, callback) {
   if (transaction && transaction.connector === this) {
     if (!transaction.connection) {
       return process.nextTick(function() {
-        callback(new Error(g.f('Connection does not exist')))
+        callback(new Error(g.f('Connection does not exist')));
       });
     }
     if (transaction.txId !== transaction.connection.txId) {
       return process.nextTick(function() {
-        callback(new Error(g.f('Transaction is not active')))
+        callback(new Error(g.f('Transaction is not active')));
       });
     }
     debug('Execute SQL within a transaction');
--- a/lib/postgresql.js
+++ b/lib/postgresql.js
@@ -194,8 +194,13 @@ PostgreSQL.prototype.executeSQL = function(sql, params, options, callback) {
   }
 
   var transaction = options.transaction;
-  if (transaction && transaction.connection &&
-    transaction.connector === this) {
+  if (transaction && transaction.connector === this) {
+    if (!transaction.connection) {
+      return callback(new Error(g.f('Connection does not exist')));
+    }
+    if (transaction.txId !== transaction.connection.txId) {
+      return callback(new Error(g.f('Transaction is not active')));
+    }
     debug('Execute SQL within a transaction');
     // Do not release the connection
     executeWithConnection(transaction.connection, null);
--- a/lib/postgresql.js
+++ b/lib/postgresql.js
@@ -196,10 +196,14 @@ PostgreSQL.prototype.executeSQL = function(sql, params, options, callback) {
   var transaction = options.transaction;
   if (transaction && transaction.connector === this) {
     if (!transaction.connection) {
-      return callback(new Error(g.f('Connection does not exist')));
+      return process.nextTick(function() {
+        callback(new Error(g.f('Connection does not exist')))
+      });
     }
     if (transaction.txId !== transaction.connection.txId) {
-      return callback(new Error(g.f('Transaction is not active')));
+      return process.nextTick(function() {
+        callback(new Error(g.f('Transaction is not active')))
+      });
     }
     debug('Execute SQL within a transaction');
     // Do not release the connection
--- a/lib/transaction.js
+++ b/lib/transaction.js
@@ -5,6 +5,8 @@
 
 'use strict';
 var debug = require('debug')('loopback:connector:postgresql:transaction');
+var uuid = require('uuid');
+var Transaction = require('loopback-connector').Transaction;
 
 module.exports = mixinTransaction;
 
@@ -18,6 +20,7 @@ function mixinTransaction(PostgreSQL) {
    * @param cb
    */
   PostgreSQL.prototype.beginTransaction = function(isolationLevel, cb) {
+    var connector = this;
     debug('Begin a transaction with isolation level: %s', isolationLevel);
     this.pg.connect(function(err, connection, done) {
       if (err) return cb(err);
@@ -25,7 +28,10 @@ function mixinTransaction(PostgreSQL) {
       connection.query('BEGIN TRANSACTION ISOLATION LEVEL ' + isolationLevel,
         function(err) {
           if (err) return cb(err);
-          cb(null, connection);
+          var tx = new Transaction(connector, connection);
+          tx.txId = uuid.v1();
+          connection.txId = tx.txId;
+          cb(null, tx);
         });
     });
   };
@@ -65,6 +71,7 @@ function mixinTransaction(PostgreSQL) {
 
   PostgreSQL.prototype.releaseConnection = function(connection, err) {
     if (typeof connection.autorelease === 'function') {
+      connection.txId = null;
       connection.autorelease(err);
       connection.autorelease = null;
     } else {
--- a/package.json
+++ b/package.json
@@ -25,7 +25,8 @@
     "debug": "^2.1.1",
     "loopback-connector": "^4.0.0",
     "pg": "^6.0.0",
-    "strong-globalize": "^2.6.2"
+    "strong-globalize": "^2.6.2",
+    "uuid": "^3.0.1"
   },
   "devDependencies": {
     "eslint": "^2.13.1",
--- a/test/init.js
+++ b/test/init.js
@@ -29,6 +29,7 @@ if (process.env.CI) {
       'emptytest',
     username: process.env.PGUSER,
     password: process.env.PGPASSWORD,
+    poolSize: 100,
   };
 }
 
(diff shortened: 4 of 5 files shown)


slnode:

slnode:

slnode:

zbarbuto:
@superkhau @raymondfeng @jannyHou

zbarbuto:
I suppose the id need not be a uuid. Could easily be an incrementing integer. Open to opinions.

zbarbuto:
Not sure what was happening with the failed tests on node 4.x and 6.x - all passing here on my end 7.x but have fixed the eslint errors in the test files.

zbarbuto:
@jannyHou Any idea what might be causing the failures? Is it something I can fix up?

zbarbuto:
Seems to be unrelated to this PR:

>transactions bulk.should work with bulk transactions
> sorry, too many clients already

What is the connection limit on the build server's postgres?

lib/transaction.js
```diff
@@ -18,14 +20,18 @@ function mixinTransaction(PostgreSQL) {
...
       if (err) return cb(err);
       connection.autorelease = done;
       connection.query('BEGIN TRANSACTION ISOLATION LEVEL ' + isolationLevel,
         function(err) {
           if (err) return cb(err);
-          cb(null, connection);
+          var tx = new Transaction(connector, connection);
+          tx.txId = uuid.v1();
+          connection.txId = tx.txId;
+          cb(null, tx);
```
jannyHou: I think here we should callback `connection` instead of `tx`?

jannyHou:
@slnode test please

jannyHou:
@zbarbuto In my opinion the expected behaviour is: 
If we do `tx.commit()`, then use that expired `tx` in other operations like `create`, postgresql db should return us an error. 
**NOT**
We add a `uuid` to check the transaction still active on our end

So if the example use case you described in the issue happens, it looks like a bug of nodejs postgresql driver to me :(

@raymondfeng any thought?

lib/transaction.js
zbarbuto: The goal here is to make the new connection and transaction and say "this connection is currently working on this transaction" allowing for the error if you pass in a transaction with a connection id that doesn't match. 

If you callback with connection, loop back connector will make a transaction out of it and pas it up the chain. This means you how no tie between the id of the connection and the current transaction.

zbarbuto:
I don't think it's a problem with the driver since the driver doesn't actually handle transactions natively. You're just executing normal transaction sql on a single connection:

https://github.com/brianc/node-postgres/wiki/Transactions

This assumes you're managing the connection in such a way that no other operations can execute on it until the transaction has resolved

In the case of the loop back connector, the connection gets released from the transaction but nothing notifies an ongoing promise chain that the connection being used is no longer part of a transaction. If it doesn't error out creates, they just end up in the database which is very dangerous and can result in orphaned records:


```javascript 
ModelA.create(tx)
.then(() => ModelB.create(tx)) // Tx times out here
.then(() => JoinTable.create({ model a, model b, tx})
```

If the transaction rolls back, a and b don't his the database (they were part of the transaction) but the join between them does (it was on a connection that Lo longer has an active transaction). Surely this can't be expected behaviour.

jannyHou:
@slnode test please

jannyHou:
@zbarbuto I think the default poolSize is 10, see https://github.com/brianc/node-postgres/blob/master/lib/defaults.js#L40, while increasing the limit doesn't solve the issue but make it worse...

zbarbuto:
Thanks for trying. The test that is failing is one that I added a little while ago so strange that it starts failing now. 

Just waiting on approval from @raymondfeng?

Did you want me to switch from uuid to an incrementing value to avoid unnecessary dependencies?

lib/postgresql.js
```diff
@@ -161,8 +161,13 @@ PostgreSQL.prototype.executeSQL = function(sql, params, options, callback) {
   }
 
   var transaction = options.transaction;
-  if (transaction && transaction.connection &&
-    transaction.connector === this) {
+  if (transaction && transaction.connector === this) {
+    if (!transaction.connection) {
+      return callback(new Error(g.f('Connection does not exist')));
```
raymondfeng: Please wrap it into `process.nextTick` or `setImmediate`.

lib/postgresql.js
```diff
@@ -161,8 +161,13 @@ PostgreSQL.prototype.executeSQL = function(sql, params, options, callback) {
...
+    }
+    if (transaction.txId !== transaction.connection.txId) {
+      return callback(new Error(g.f('Transaction is not active')));
```
raymondfeng: Please wrap it into `process.nextTick` or `setImmediate`.

lib/postgresql.js
zbarbuto: Done

lib/postgresql.js
zbarbuto: Done

zbarbuto:
No further ideas on why the tests might be failing?

jannyHou:
@zbarbuto I am making time to debug why `done()` is called multiple times, and why it complains too many connections....

jannyHou:
@slnode test please

lib/transaction.js
jannyHou: @zbarbuto after I fixed `if(err) return done(err)`, now we get the real error:
```shell
 1) transactions bulk should work with bulk transactions:
     Error: read ECONNRESET
      at exports._errnoException (util.js:907:11)
      at TCP.onread (net.js:558:26)
```
I suspect it's because here we callback the `tx`, then when we do `tx.commit` to release it, but we don't close/release the `connection` in L25...
The old code callback `connection` so it doesn't have this error.

jannyHou:
@slnode test please

lib/postgresql.js
```diff
@@ -161,8 +161,17 @@ PostgreSQL.prototype.executeSQL = function(sql, params, options, callback) {
   }
 
   var transaction = options.transaction;
-  if (transaction && transaction.connection &&
-    transaction.connector === this) {
+  if (transaction && transaction.connector === this) {
+    if (!transaction.connection) {
+      return process.nextTick(function() {
+        callback(new Error(g.f('Connection does not exist')))
```
jannyHou: missing `;`

lib/postgresql.js
```diff
@@ -161,8 +161,17 @@ PostgreSQL.prototype.executeSQL = function(sql, params, options, callback) {
...
+      });
+    }
+    if (transaction.txId !== transaction.connection.txId) {
+      return process.nextTick(function() {
+        callback(new Error(g.f('Transaction is not active')))
```
jannyHou: missing `;`

lib/transaction.js
zbarbuto: I don't think this is the case, as if you look at `loopback-conenctor` the commit method passes the connection of the transaction to the connector:

https://github.com/strongloop/loopback-connector/blob/master/lib/transaction.js#L48

And moreover, if you debug the code at line 76, you can see that the connection that is passed in to `releaseConnection` is definitely the connection that has `autorelease()` attached to it.

Calling back with transaction instead of the connection should make essentially no difference, as `loopback-connector` will immediately turn it into a `Transaction` if it is not already an instance of `Transaction`:

https://github.com/strongloop/loopback-connector/blob/master/lib/transaction.js#L88

I simply skip this step so we can attach the id to both the connection and the transaction.

lib/postgresql.js
jannyHou: missing `;`

lib/postgresql.js
jannyHou: missing `;`

lib/postgresql.js
zbarbuto: Fixed

lib/postgresql.js
zbarbuto: Fixed

kjdelisle:
@slnode test please

jannyHou:
@slnode test please

jannyHou:
The same failure also happens on master branch, it fails on only one machine and I set up a windows vm using same environment, but could not reproduce the error.

I will create another PR just debug why it fails. Sorry for the delay of checking this pr in, it looks good but we have to fix the test :)

zbarbuto:
No problems. Thanks for your help @jannyHou

zbarbuto:
@dhr00v Wh would the callback not be executed if `ROLLBACK` was? Could you put together an example repo for an instance where this might be the case?

sahilarora946:
@zbarbuto After the `ROLLBACK` was executed, callback will be sent to the event loop to be executed. Event loop can pick any of the available callbacks to execute. It can pick another function that can do database calls in the same transaction and because `releaseConnection` has not been called, the statement will execute.

zbarbuto:
Hmm - I see what you're saying. Do you have a proposed solution for this? Would you be willing to submit a PR?

sahilarora946:
Ok. I will submit a PR for this @zbarbuto

dhmlau:
Fixed in https://github.com/strongloop/loopback-connector-postgresql/pull/330, which is released as `loopback-connector-postgresql@3.3.2`.

Issue #231, opened 2017-03-24
Default for minimum pool size: After 10 Errors DB stops working


bobbysmith007:
### Bug or feature request

- [x] Bug
- [x] Feature request

### Description of feature (or steps to reproduce if bug)

After 10 sql errors, the application silently stops running any database code and will not execute sql any further with no indication of whats wrong. This occurs because without a minimum number of workers, the pool will wait till a worker is available, but never create any on its own.

Adding "min":3 to the datasource configuration (which is passed to pg-pool, which is passed to generic-pool) resolves this problem by assuring there are always at least 3  workers

### Link to sample repo to reproduce issue (if bug)

https://github.com/bobbysmith007/loopback-error-reproductions
`npm install .`
`npm run-script show-errors`

### Expected result

Running 15sql queries with errors from each

### Actual result (if bug)

Running 10sql queries with it silently never running the rest

### Additional information (Node.js version, LoopBack version, etc)

 * loopback: 3.4.0
 * nodejs :7.6.0

ssh24:
@bobbysmith007 If you have a fix in place, would you mind creating a pull request?

ssh24:
Should be fixed with https://github.com/strongloop/loopback-connector-postgresql/commit/0697d874083c6f13d17e6c4dd379c39ce5eef8c7. 

Closing this issue. Feel free to reopen if needed.

PR #234, opened 2017-03-29
Monkey patch the generic-pool to continue working through errors

bobbysmith007:
### Description

 * previously without a minimum the pool would stop handling database
   requests due to worker exhaustion (for any requests started before
   worker exhaustion)
 * This patch ensures that if we have waiting work, we will continue
   processing it
 * a test that fails before this patch and succeeds after this patch:
  'postgresql connector errors'

#### Related issues

#231 Default for minimum pool size: After 10 Errors DB stops working 

### Checklist

### DIFF
--- a/lib/postgresql.js
+++ b/lib/postgresql.js
@@ -50,6 +50,38 @@ exports.initialize = function initializeDataSource(dataSource, callback) {
   }
 };
 
+// there is a bug in the generic pool where it will not recreate
+// destroyed workers (even if there is waiting work to do) unless
+// there is a min specified. Make sure we keep some connections
+// SEE: https://github.com/coopernurse/node-pool/pull/186
+// SEE: https://github.com/brianc/node-pg-pool/issues/48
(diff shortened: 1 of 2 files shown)


slnode:

slnode:

slnode:

slnode:

raymondfeng:
ok to test

raymondfeng:
In the case of SQLException, do we have to tear down the connection? Maybe we should change the code so that it won't trigger the recycle of the connection?

raymondfeng:
ok to test

bobbysmith007:
I cant make heads or tails of the output above.  Is this something I need to worry about?  Is there more I need to do here?

raymondfeng:
I'm running tests to verify the pr

raymondfeng:
Merged via https://github.com/strongloop/loopback-connector-postgresql/commit/0697d874083c6f13d17e6c4dd379c39ce5eef8c7. Thanks!

PR #235, opened 2017-03-30
Fix the escape char

raymondfeng:
### Description

The current code fails the tests due to invalid escape chars.

#### Related issues

- None

### Checklist

### DIFF
--- a/lib/postgresql.js
+++ b/lib/postgresql.js
@@ -387,13 +387,13 @@ PostgreSQL.prototype.buildExpression = function(columnName, operator,
       return new ParameterizedSQL(columnName + " LIKE ? ESCAPE E'\\\\'",
           [operatorValue]);
     case 'ilike':
-      return new ParameterizedSQL(columnName + " ILIKE ? ESCAPE '\\'",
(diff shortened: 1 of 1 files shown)


lib/postgresql.js
```diff
@@ -387,13 +387,13 @@ PostgreSQL.prototype.buildExpression = function(columnName, operator,
       return new ParameterizedSQL(columnName + " LIKE ? ESCAPE E'\\\\'",
           [operatorValue]);
     case 'ilike':
-      return new ParameterizedSQL(columnName + " ILIKE ? ESCAPE '\\'",
+      return new ParameterizedSQL(columnName + " ILIKE ? ESCAPE '\\\\'",
```
jannyHou: `" LIKE ? ESCAPE E'\\\\'"` the case above has an extra `E` before 4 back-slash, should it be added here as well? The change LGTM while CI fails a lot.

And the transaction test fails due to https://github.com/strongloop/loopback-connector-postgresql/pull/230/commits/2876a95a504451bae68b4fc6aa783ecb806bf9f9
we can fix the test case in this pr.

raymondfeng: Added `E` to make it consistent

PR #244, opened 2017-04-12
Reuse the data source to avoid too many clients

raymondfeng:
### Description

A lot of tests from juggler calls `getDataSource/getSchema` to set up the DB connections without disconnecting them. As a result, postgresql connector tests run out of client connections. This PR allows the reuse of ds.

#### Related issues

- None

### Checklist

### DIFF
--- a/test/init.js
+++ b/test/init.js
@@ -43,9 +43,14 @@ global.getDBConfig = function(useUrl) {
   };
   return settings;
 };
+
+var db;
 global.getDataSource = global.getSchema = function(useUrl) {
+  // Return cached data source if possible to avoid too many client error
+  // due to multiple instances of connection pools
+  if (!useUrl && db) return db;
   var settings = getDBConfig(useUrl);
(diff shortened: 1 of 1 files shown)


kjdelisle:
@raymondfeng You're okay to merge this.

test/init.js
```diff
@@ -43,9 +43,14 @@ global.getDBConfig = function(useUrl) {
   };
   return settings;
 };
+
+var db;
 global.getDataSource = global.getSchema = function(useUrl) {
+  // Return cached data source if possible to avoid too many client error
+  // due to multiple instances of connection pools
+  if (!useUrl && db) return db;
```
jannyHou: I tried to clear the models attached to db in previous tests
```
db.connector._models = {}
```
But it breaks some test like https://github.com/strongloop/loopback-datasource-juggler/blob/master/test/basic-querying.test.js#L60
which runs getSchema() to get db again for `unknownid` generator. Will fix it then.

PR #245, opened 2017-04-13
dbdefaults: Cleanup InvalidDefault def after test

kjdelisle:
### Description

After switching to a single DataSource instance for the tests, the
InvalidDefault definition remains in the connector's _models
collection. Removed this definition to allow use of db.automigrate
in subsequent tests.


kjdelisle:
@slnode test please

Issue #247, opened 2017-04-19
typecast date/time types to text in query


ptrblgh:
I have a datatable and a text input filed. The expected behaviour would be to perform a search on postgresql table fields (including the timestamp ones, in this example: stamp).
I use loopback's where filter's like operator. But unfortunetly a cast would be necessary on the table's field (stamp::TEXT) to make it work.

the model's property:
```json
"stamp": {
      "type": "String",
      "required": true,
      "length": null,
      "precision": null,
      "scale": null,
      "postgresql": {
        "columnName": "stamp",
        "dataType": "timestamp with time zone",
        "dataLength": null,
        "dataPrecision": null,
        "dataScale": null,
        "nullable": "NO"
      }
    },
```

the url:
`http://localhost:3000/api/users?filter[where][stamp][like]=2017`

the error:
```
length: 213
severity: ERROR
code: 42883
hint: No operator matches the given name and argument type(s). You might need to add explicit type casts.
position: 246
file: parse_oper.c
line: 766
routine: op_error
error: operator does not exist: timestamp with time zone ~~ text
    at Connection.parseE (...node_modules\pg\lib\connection.js:569:11)
    at Connection.parseMessage (...node_modules\pg\lib\connection.js:396:17)
    at Socket. (..node_modules\pg\lib\connection.js:132:22)
    at emitOne (events.js:96:13)
    at Socket.emit (events.js:188:7)
    at readableAddChunk (_stream_readable.js:176:18)
    at Socket.Readable.push (_stream_readable.js:134:10)
    at TCP.onread (net.js:548:20)
```

the working query would be:
`select stamp from du_user where stamp::text like '%2017%'`

Right now i don't have any idea how to achieve a query like that with loopback/loopback-connector-postgresql. 
Any suggestions/help would be appreciated.

ptrblgh:
I think i have found a solution.

As the postgres documentations says: 
>The LIKE expression returns true if the ***string*** is contained in the set of strings represented by pattern.

Therefore typecasting to text explicitly the target column not result any error.

The following lines should be changed:
https://github.com/strongloop/loopback-connector-postgresql/blob/master/lib/postgresql.js#L429
https://github.com/strongloop/loopback-connector-postgresql/blob/master/lib/postgresql.js#L432
[...]

PR #255, opened 2017-05-06
add explicit typecasts to pattern matching

ptrblgh:
Related to #247 .

### DIFF
--- a/test/postgresql.test.js
+++ b/test/postgresql.test.js
@@ -403,7 +403,7 @@ describe('postgresql connector', function() {
     it('supports nlike for no match for date types (with explicit typecast)', function(done) {
       PostWithDate.find({where: {content: {nlike: '%07%'}}},
         function(err, posts) {
-          if (err) return done(err);
+          should.not.exists(err);
           posts.length.should.equal(2);
           done();
         });
--- a/test/postgresql.test.js
+++ b/test/postgresql.test.js
@@ -418,41 +418,47 @@ describe('postgresql connector', function() {
     });
 
     function deleteTestFixtures(done) {
-      Post.destroyAll(done);
+      Post.destroyAll(function(err) {
+        should.not.exist(err);
+        if (PostWithDate) PostWithDate.destroyAll(done);
+        done();
+      });
     }
 
     function createTestFixtures(done) {
+      PostWithDate = db.define('PostWithDate', {
+        title: {type: String, length: 255, index: true},
+        content: {type: String},
+        created: {
+          type: String,
+          postgresql: {
+            dataType: 'timestamp with time zone',
+          },
+        },
+      });
       db.automigrate(function(err, result) {
         if (err) throw err;
-        PostWithDate = db.define('PostWithDate', {
-          title: {type: String, length: 255, index: true},
-          content: {type: String},
-          created: {
-            type: String,
-            postgresql: {
-              dataType: 'timestamp with time zone',
-            },
-          },
-        });
         Post.create([{
           title: 't1',
           content: 'T1_TestCase',
         }, {
           title: 't2',
           content: 'T2_TheOtherCase',
-        }]);
-        PostWithDate.create([
-          {
-            title: 'Title 1',
-            content: 'Content 1',
-            created: '2017-05-17 12:00:01',
-          },
-          {
-            title: 'Title 2',
-            content: 'Content 2',
-            created: '2017-04-17 12:00:01',
-          },
-        ], done);
+        }], function(err, result) {
+          should.not.exist(err);
+          PostWithDate.create([
+            {
+              title: 'Title 1',
+              content: 'Content 1',
+              created: '2017-05-17 12:00:01',
+            },
+            {
+              title: 'Title 2',
+              content: 'Content 2',
+              created: '2017-04-17 12:00:01',
+            },
+          ], done);
+        });
       });
     }
   });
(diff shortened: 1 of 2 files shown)


slnode:

slnode:

slnode:

slnode:

ssh24:
@ptrblgh Thank you for the contribution. Could you also add some unit tests to prevent future regression?

Also please sign the [`Contributor License Agreement`](https://cla.strongloop.com/agreements/strongloop/loopback-connector-postgresql) before we can proceed with the PR.

ssh24:
@ptrblgh Any updates on adding some test cases with this patch?

ptrblgh:
Not yet, I'm stuck right now to make the test fail but working on it.

ssh24:
Not sure what you mean. Are you having difficulties writing the test case?

Here is a quick example of one case:
```javascript
  postgresDs.automigrate(function(err, result) {
    if (err) throw err;
    Expense.create([
      {
        Description: 'Expense 1',
        Amount: 12,
        Stamp: '1999-01-08 12:25:00',
      },
    ], function(err, result) {
      if (err) throw err;
      Expense.find({where: {Stamp: {like: '1999%'}}}, function(err, result) {
        if (err) throw err;
        console.log('Result: ' + JSON.stringify(result));
      });
    });
```

Please note this is just the use case, you need to have `asserts` to make it a more of a test case. Also please be sure to add test cases where the operation is of type `ilike`, `nlike`, `nilike`.

ssh24:
@ptrblgh Any updates on adding the test cases?

ptrblgh:
@ssh24 Sorry for the delay. I have added tests for this patch.

ssh24:
@ptrblgh Thank you. I will be looking into it soon. Could you please rebase your branch and fix your commit linter?

Rules for commit messages:
- The first line of the commit message should be no longer than 50 characters long
- The second line should be empty
- Every subsequent line should be no longer than 72 characters long

You could also give me access to handle the rebase/commit fixing by clicking `Allow edit from Maintainers` on the right hand side of this page.

ptrblgh:
No problem, allow granted. 👍

ssh24:
Thank you. Code rebased.

Also PR looks good to me. Brilliant! 👍 

@slnode test please

test/postgresql.test.js
```diff
@@ -384,18 +384,76 @@ describe('postgresql connector', function() {
...
+              }
+            });
+        Post.create([{
+          title: 't1',
+          content: 'T1_TestCase',
+        }, {
+          title: 't2',
+          content: 'T2_TheOtherCase',
+        }]);
+        PostWithDate.create([
```
ssh24: @ptrblgh There is an issue here. Since the functions are async, it does not work like that and hence the tests are failing. You need to do something like this:
```
Post.create([...], function(err, result) { should.not.exist(err); PostWithDate.create([..], done});
```

ptrblgh:
@ssh24 Thank you for the hint! I have changed the code, but can't leave the `PostWithDate` definition inside automigrate, because the test will fail with missing `postwithdate` table relation.

I came up with this solution (tests are OK):
```js
function createTestFixtures(done) {
      PostWithDate = db.define('PostWithDate', {
        title: {type: String, length: 255, index: true},
        content: {type: String},
        created: {
          type: String,
          postgresql: {
            dataType: 'timestamp with time zone'
          }
        }
      });
      db.automigrate(function(err, result) {
        if (err) throw err;
        Post.create([{
          title: 't1',
          content: 'T1_TestCase',
        }, {
          title: 't2',
          content: 'T2_TheOtherCase',
        }], function(err, result) {
          should.not.exist(err);
          PostWithDate.create([
            {
              title: 'Title 1',
              content: 'Content 1',
              created: '2017-05-17 12:00:01'
            },
            {
              title: 'Title 2',
              content: 'Content 2',
              created: '2017-04-17 12:00:01'
            },
          ], done);
        });
      });
    }
```
If this code looks good for you I'll make a commit :).

Other question:
I got 2 other tests failing (not connected for this one):
  1) manipulation updateOrCreate when forceId is true fails when id does not exist in db & validate is false:
     Uncaught TypeError: Cannot read property 'statusCode' of null
 2) manipulation updateOrCreate when forceId is true fails when id does not exist in db & validate is false when using updateAttributes:
     Uncaught TypeError: Cannot read property 'statusCode' of null

Is this only for me? Should I change something in my config to pass these tests too?

test/postgresql.test.js
```diff
@@ -384,18 +384,76 @@ describe('postgresql connector', function() {
...
+    it('supports case insensitive queries using nilike for date types (with explicit typecast)', function(done) {
+      PostWithDate.find({where: {created: {nilike: '%07%'}}}, function(err, result) {
+        should.not.exists(err);
+        result.length.should.equal(2);
+        done();
+      });
+    });
+
     function deleteTestFixtures(done) {
       Post.destroyAll(done);
```
ssh24: Could you also destroy the data after the test is done here? 

```javascript
    function deleteTestFixtures(done) {
      Post.destroyAll(function(err) {
        should.not.exist(err);
        PostWithDate.destroyAll(done);
      });
    }
```

ptrblgh:
@ssh24  clean up added to the test case.

ssh24:
@ptrblgh Thank you. Rebased your branch. Code looks good to me. 👍 

@slnode test please

Issue #58, opened 2015-02-05
Numeric is converted to String


bitmage:
With this in my model.json:

``` json
{
  "amount": {
    "type": "number",
    "human": "Amount",
    "postgresql": {
      "dataType": "numeric",
      "dataPrecision": 16,
      "dataScale": 4
    }
  }
}
```

When I store the value `50000.5413` I get back `'50000.5413'`.

kelchm:
I'm also seeing this issue -- any update?

mirow:
The problem (and solution) is here: 
https://github.com/brianc/node-pg-types/issues/28

ssh24:
Added docs to allow the users know of this behaviour. Closing this issue.

PR #265, opened 2017-06-22
Add docs on numeric data type

ssh24:
### Description
The node.js driver for postgres returns a `Numeric` type value as a **string** to avoid data precision loss. Since we as the ORM simply just return back the result from the driver, we need to document this behaviour.

connect to https://github.com/strongloop/loopback-connector-postgresql/issues/58

### DIFF
--- a/README.md
+++ b/README.md
@@ -363,6 +363,12 @@ See [LoopBack types](http://loopback.io/doc/en/lb3/LoopBack-types.html) for de
   </tbody>
 </table>
 
+## Numeric Data Type
+
+**Note**: The [node.js driver for postgres](https://github.com/brianc/node-postgres) by default casts `Numeric` type as a string on `GET` operation. This is to avoid _data precision loss_ since `Numeric` types in postgres cannot be safely converted to JavaScript `Number`.
+
+For details, see the corresponding [driver issue](https://github.com/brianc/node-pg-types/issues/28).
+
 ## Discovery and auto-migration
 
 ### Model discovery


README.md
```diff
@@ -363,6 +363,12 @@ See [LoopBack types](http://loopback.io/doc/en/lb3/LoopBack-types.html) for de
   </tbody>
 </table>
 
+## Numeric Data Type
+
+**Note**: Please be advised the [node.js driver for postgres](https://github.com/brianc/node-postgres) by default casts `Numeric` type as a string on `GET` operation. This is to avoid _data precision loss_ since `Numeric` types in postgres cannot be safely converted to JavaScript `Number`. 
+
+[Link to the corresponding driver issue](https://github.com/brianc/node-pg-types/issues/28).
```
dhmlau: maybe "For details, see the corresponding driver issue"?

README.md
```diff
@@ -363,6 +363,12 @@ See [LoopBack types](http://loopback.io/doc/en/lb3/LoopBack-types.html) for de
   </tbody>
 </table>
 
+## Numeric Data Type
+
+**Note**: Please be advised the [node.js driver for postgres](https://github.com/brianc/node-postgres) by default casts `Numeric` type as a string on `GET` operation. This is to avoid _data precision loss_ since `Numeric` types in postgres cannot be safely converted to JavaScript `Number`.
```
dhmlau: I don't think we need the "please be advised"..  but I'm not english expert :)

Issue #258, opened 2017-05-15
Transactions are broken on v3.0.1


StevenWinde:
# Description/Steps to reproduce
This works fine in 3.0.0 but not in 3.0.1
```
this.beginTransaction( 'READ COMMITTED', ( err, tx )=>{
  this.updateAll( { foo: 'bar' },  { foo: 'bar2'},  { transaction: tx },
  ( err )=>{
    // err is Transaction is not active.
  }
}
```

cadesalaberry:
Version 3.0.1 is also not working for me...

Do not use this version in production !

```
Error: Transaction is not active
at /opt/app/node_modules/loopback-connector-postgresql/lib/postgresql.js:205:18
at nextTickCallbackWith0Args (node.js:420:9)
at process._tickDomainCallback (node.js:390:13)
From previous event:
  at Object.createPromiseCallback (/opt/app/node_modules/loopback-datasource-juggler/lib/utils.js:492:17)
  at Function.DataAccessObject.update.DataAccessObject.updateAll (/opt/app/node_modules/loopback-datasource-juggler/lib/dao.js:2551:20)
```

jmls:
_really_ ? transactions are broken, and no-one from SL responds ? It's this sort of thing that really makes me worry about strongloop. Does anyone have a fix ?

cadesalaberry:
@bajtos could you have a look at it please ?

bajtos:
Hello, I am not familiar with PostgreSQL myself. The connector version 3.0.1 contains the following changes (see https://github.com/strongloop/loopback-connector-postgresql/releases/tag/v3.0.1):

> * Add docker setup (#256) (Sakib Hasan)
> * Allow explicit numeric datatype (#254) (Sakib Hasan)
> * Allow non-id serial properties (#198) (zbarbuto)
> * Revert PR #246 (#248) (Sakib Hasan)
> * Add loopback-connector as peer dependencies (#246) (Russ Tyndall)
> * Fix operations on ended transactions (zbarbuto)
> * dbdefaults: Cleanup InvalidDefault def after test (Kevin Delisle)
>  * Reuse the data source to avoid too many clients (Raymond Feng)

I guess https://github.com/strongloop/loopback-connector-postgresql/pull/230 may be the cause of this regression?

@raymondfeng @jannyHou @zbarbuto @kjdelisle could you PTAL?

kjdelisle:
[...]

Issue #208, opened 2017-02-08
Deadlock possible when using transactions in bulk


zbarbuto:
### Bug or feature request

- [x] Bug
- [ ] Feature request

### Description of feature (or steps to reproduce if bug)

When transactions are created, they are built on the same pool as normal requests. It is possible to get into a deadlocked state if there are enough open transactions waiting for side effects (like `.observe` model hooks).

For example, if our model has a `'loaded'` hook on it that calls `find` on another model before calling `next()`, and then we call the following with a large array of instances of the model in `modelArray`:

`return Promise.all(modelArray.map(model => model.updateAttribute('attr', 'val', {transaction: tx}));`

The server will deadlock while it the `'loaded'` hook waits for a free connection to perform its find (which can't happen since all the connections in the pool are tied up by the open transactions).

### Expected outcome

I suggest that the transaction pool be kept separate and leave at least one connection always free to allow non-transaction side effects to be completed.

zbarbuto:
Worth nothing that while this is technically a programmer error for something like a before save hook (since you can tap into the transaction and use it if it exits), in instances like loaded and after save I don't think it's possible to get at the transaction so I feel it would be better if the connector also safeguarded you just in case.

PR #268, opened 2017-06-26
Fix tranaction operations when called from the model

zbarbuto:
### Description

I was unable to reproduce #258 exactly (when I hear back I will look into this further) but I did find issues when transaction operations where called when Transaction is called from the model.

When called directly on the model, commit and rollback do not go through `transaction.js` in `loopback-connector` so what the Postgresql connector is the `Transaction` rather than the `connection`. This PR checks for that case.

#### Related issues

- #258 

connect to https://github.com/strongloop/loopback-connector-postgresql/issues/258

### Checklist

### DIFF
--- a/lib/transaction.js
+++ b/lib/transaction.js
@@ -27,9 +27,10 @@ function mixinTransaction(PostgreSQL) {
       connection.query('BEGIN TRANSACTION ISOLATION LEVEL ' + isolationLevel,
         function(err) {
           if (err) return cb(err);
-          connection.txId = uuid.v1();
-          connector.activeTransactions[connection.txId] = true;
-          cb(null, connection);
+          var tx = new Transaction(connector, connection);
+          tx.txId = uuid.v1();
+          connection.txId = tx.txId;
+          cb(null, tx);
         });
     });
   };
@@ -42,6 +43,10 @@ function mixinTransaction(PostgreSQL) {
   PostgreSQL.prototype.commit = function(connection, cb) {
     debug('Commit a transaction');
     var self = this;
+    if (connection instanceof Transaction) {
+      // If called directly on the model rather than through Transaction
+      connection = connection.connection;
+    }
     connection.query('COMMIT', function(err) {
       self.releaseConnection(connection, err);
       cb(err);
@@ -56,6 +61,10 @@ function mixinTransaction(PostgreSQL) {
   PostgreSQL.prototype.rollback = function(connection, cb) {
     debug('Rollback a transaction');
     var self = this;
+    if (connection instanceof Transaction) {
+      // If called directly on the model rather than through Transaction
+      connection = connection.connection;
+    }
     connection.query('ROLLBACK', function(err) {
       //if there was a problem rolling back the query
       //something is seriously messed up.  Return the error
@@ -68,9 +77,7 @@ function mixinTransaction(PostgreSQL) {
   };
 
   PostgreSQL.prototype.releaseConnection = function(connection, err) {
-    var self = this;
     if (typeof connection.autorelease === 'function') {
-      delete self.activeTransactions[connection.txId];
       connection.txId = null;
       connection.autorelease(err);
       connection.autorelease = null;
--- a/lib/transaction.js
+++ b/lib/transaction.js
@@ -43,10 +43,6 @@ function mixinTransaction(PostgreSQL) {
   PostgreSQL.prototype.commit = function(connection, cb) {
     debug('Commit a transaction');
     var self = this;
-    if (connection instanceof Transaction) {
-      // If called directly on the model rather than through Transaction
-      connection = connection.connection;
-    }
     connection.query('COMMIT', function(err) {
       self.releaseConnection(connection, err);
       cb(err);
@@ -61,10 +57,6 @@ function mixinTransaction(PostgreSQL) {
   PostgreSQL.prototype.rollback = function(connection, cb) {
     debug('Rollback a transaction');
     var self = this;
-    if (connection instanceof Transaction) {
-      // If called directly on the model rather than through Transaction
-      connection = connection.connection;
-    }
     connection.query('ROLLBACK', function(err) {
       //if there was a problem rolling back the query
       //something is seriously messed up.  Return the error
--- a/lib/transaction.js
+++ b/lib/transaction.js
@@ -28,10 +28,9 @@ function mixinTransaction(PostgreSQL) {
       connection.query('BEGIN TRANSACTION ISOLATION LEVEL ' + isolationLevel,
         function(err) {
           if (err) return cb(err);
-          var tx = new Transaction(connector, connection);
-          tx.txId = uuid.v1();
-          connection.txId = tx.txId;
-          cb(null, tx);
+          connection.txId = uuid.v1();
+          connector.activeTransactions[connection.txId] = true;
+          cb(null, connection);
         });
     });
   };
@@ -44,10 +43,6 @@ function mixinTransaction(PostgreSQL) {
   PostgreSQL.prototype.commit = function(connection, cb) {
     debug('Commit a transaction');
     var self = this;
-    if (connection instanceof Transaction) {
-      // If called directly on the model rather than through Transaction
-      connection = connection.connection;
-    }
     connection.query('COMMIT', function(err) {
       self.releaseConnection(connection, err);
       cb(err);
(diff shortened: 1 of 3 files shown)


slnode:

slnode:

slnode:

zbarbuto:
@raymondfeng @jannyHou @kjdelisle @dhmlau

If anyone can add a test reproducing the original issue in #258 I'll investigate further.

dhmlau:
I can reproduce the problem by creating the following js in `boot` folder:
(Customer model with `name` property)

```
'use strict';

module.exports = function(app) {
  var customer = app.models.Customer;
  customer.beginTransaction('READ COMMITTED', (err, tx)=>{
    customer.updateAll({name: 'aaa'}, {name: 'bbb'}, {transaction: tx}, (err)=>{
      console.log('err = ', err);
    });
  });
};
```

dhmlau:
@zbarbuto  while I believe your changes might fix some other problem, I don't think it's the cause of #258.  Error was thrown before commit function.

jannyHou:
Thank you @zbarbuto I will review it asap, and some investigation @dhmlau did yesterday shows the problem is:

In postgresql connector, the `transaction` instance is created from the one in:
 **`loopback-connector-postgresql/node_modules/loopback-connector`**,
see [`var tx = new Transaction()`](https://github.com/strongloop/loopback-connector-postgresql/blob/eeaa89eec8758c07ce4c265ded1970b3a102c952/lib/transaction.js#L31)

While when the postgresql's `beginTransaction` function calls back, it goes to `begin` function in:
**`loopback-datasource-juggler/node_modules/loopback-connector`**,
```js
if (!(connection instanceof Transaction)) {
      tx = new Transaction(connector, connection);
}
```
is get executed since the `connection` is an instanceof the first `lb-connector`, not instanceof the second `lb-connector`, and when instantiate `Transaction` again, the `txId` is removed, and therefore the connector considers the transaction is not active.
see [code](https://github.com/strongloop/loopback-connector/blob/master/lib/transaction.js#L88-L90)

zbarbuto:
@dhmlau Thanks for the steps to reproduce. Can see it happen myself now and am investigating.

zbarbuto:
@jannyHou You're correct. This is caused when the version of `loopback-connector` required by the app is different to that required by `loopback-connector-postgresql` resulting in the connector's `instanceof` returning false. The real fix would be to ensure that all components are using the same version (e.g. for the postgres connector require `loopback-connector` as a peer dep _or_ avoid requiring loopback-connector at the app level). 

However, I've fixed it to avoid using `Transaction` to track transaction state and instead use an object on the connector. It's a bit of a dirty fix but it means that the issue of #230 is still handled but allows differing versions of `loopback-connector`.

Not sure the best way to add a test to avoid the regression in the future. Perhaps someone else could advise?

dhmlau:
@zbarbuto @jannyHou , the cause of the problem is because there are different definition of `Transaction` which causes this line (https://github.com/strongloop/loopback-connector/blob/master/lib/transaction.js#L88) to fail.   Would it make sense to change the if-statement from
`if (!(connection instanceof Transaction))` 
to 
`if (connection.txId === undefined)`

cc @raymondfeng

zbarbuto:
@dhmlau `connection.txId` is only added by this connector (was part of #230). Changing would probably break other sql connectors (such as MySQL)

The definition of `Transaction` is the file required in can be different because the version is different (`loopback-connector-postgresql` requires the one in its own `./node_modules` folder if the version is different from the one in the parent package's `../node_modules` which is why `instanceof` isn't working).

dhmlau:
@jannyHou  @zbarbuto , I've verified the code with and without transaction. 
@jannyHou , PTAL.

raymondfeng:
To avoid the failure of `instanceof Transaction`, we probably should use duck-typing by testing the properties such as `if (typeof transaction.txId === 'string')`. 

I'm not sure we need to use `activeTransactions` to keep track of the valid transactions. If we have to do so, we can set `connector` to `Transaction` object and test if `tx.connector === this`.

zbarbuto:
@raymondfeng But `loopback-connector` is what checks `instanceof Transaction`, not `loopback-connector-postgresql` so it does not know about `txId` - its unique to this connector (see https://github.com/strongloop/loopback-connector-postgresql/pull/268#issuecomment-311536909). The only reason `txId` was added to this connector was to avoid executing CRUD operations on a connection that originally had a transaction but has since been committed (#230).

Moving to `loopback-connector` as a peer-dependency is probably the 'cleaner' solution as it doesn't require any code changes but is potentially more breaking.

dhmlau:
It looks like only `Transaction` has `connector`.  How about we check:
`if (connection.connector === undefined)`
in this line https://github.com/strongloop/loopback-connector/blob/master/lib/transaction.js#L88

raymondfeng:
@dhmlau We can probably fix the code at https://github.com/strongloop/loopback-connector/blob/master/lib/transaction.js#L88 as follows:
```
    var tx = connection;
    if (tx.connector !== this) {
      // The connector returns a connection instead of transaction
      tx = new Transaction(connector, connection);
    } else {
      tx = new Transaction(tx.connector, tx.connection);
    }
```

@dhmlau Can you verify?

dhmlau:
Verified.  Works with and without transaction.  
@zbarbuto , are you ok with making the change suggested by @raymondfeng?

dhmlau:
@zbarbuto, do you have a chance to review the suggested changes?  Thanks!

dhmlau:
Apologize that there has been some environment issue, therefore, the changes were not got reflected when I was doing the testing.  I will have to do the verification again. 
If only changes in `loopback-connector` is needed, I'll create a separate PR.  Thanks!

zbarbuto:
@dhmlau I'll PR the suggested change to `loopback-connector` shortly.

zbarbuto:
@dhmlau Update: I just tested with @raymondfeng 's suggested change and tests start failing after the bulk transactions (database deadlocks) see #208.

https://github.com/strongloop/loopback-connector-postgresql/blob/master/test/postgresql.transaction.test.js#L62

dhmlau:
right.  there's some problem with the suggested change.  @jannyHou proposed something else, let me try it out and let you know.

dhmlau:
@zbarbuto , I put @jannyHou 's proposed fix in https://github.com/strongloop/loopback-connector/pull/107.  However, currently there's some issue with the downstream `loopback-connector-dashdb` CI.  We're working on that.

dhmlau:
@zbarbuto, we've implemented a fix in `loopback-connector` which seems to be a better fix than patching it on the postgreql repo, and also include the test cases in there.   https://github.com/strongloop/loopback-connector/pull/109

I think the tests that you've added in `postgresql.transaction.test.js` would be useful.  If you agree, could you please remove the changes in lib\*.js and incorporate the review comments in the test.js?  Thanks a lot!

zbarbuto:
@dhmlau  Updated tests, removed active transaction tracking and rebased off latest master.

lib/transaction.js
```diff
@@ -44,6 +43,10 @@ function mixinTransaction(PostgreSQL) {
   PostgreSQL.prototype.commit = function(connection, cb) {
     debug('Commit a transaction');
     var self = this;
+    if (connection instanceof Transaction) {
```
dhmlau: I've tested our changes in `loopback-connector` and it works without any changes in this repo. 
Any reason we need to make this change?

lib/transaction.js
```diff
@@ -58,6 +61,10 @@ function mixinTransaction(PostgreSQL) {
   PostgreSQL.prototype.rollback = function(connection, cb) {
     debug('Rollback a transaction');
     var self = this;
+    if (connection instanceof Transaction) {
```
dhmlau: same as above.

lib/transaction.js
dhmlau: Actually i think CI failed because of this.

zbarbuto: Missed this in the revert. Have removed now.

lib/transaction.js
zbarbuto: as above

zbarbuto:
Rebased again.

dhmlau:
@slnode  test please

lib/transaction.js
```diff
@@ -6,7 +6,6 @@
 'use strict';
 var debug = require('debug')('loopback:connector:postgresql:transaction');
 var uuid = require('uuid');
-var Transaction = require('loopback-connector').Transaction;
```
dhmlau: the last one.. i think this is needed.

zbarbuto:
Done

dhmlau:
Thanks for going through all this with us!  Your PR has landed!

PR #273, opened 2017-07-10
Test/transaction

jannyHou:
### Description

Test case for loopback-connector PR https://github.com/strongloop/loopback-connector/pull/107

### DIFF
--- a/test/postgresql.transaction.test.js
+++ b/test/postgresql.transaction.test.js
@@ -127,7 +127,7 @@ describe('transactions', function() {
           function(err, changes) {
             tx.commit(function(err) {
               if (err) {
-                done(err);
+                return done(err);
               }
               done();
             });
--- a/test/postgresql.transaction.test.js
+++ b/test/postgresql.transaction.test.js
@@ -109,6 +109,33 @@ describe('transactions', function() {
     it('should see the committed insert', expectToFindPosts(post, 1));
   });
 
+  describe('update all', function() {
+    var p1Content = {title: 'p1', content: 'post-a'};
+    var p2Content = {title: 'p2', content: 'post-a'};
+
+    before(function(done) {
+      Post.create(p1Content, function(err, p1) {
+        Post.create(p1Content, function(err, p2) {
+          done();
+        });
+      });
+    });
+
+    it('should work with update all', function(done) {
+      Transaction.begin(db.connector, Transaction.READ_COMMITTED, function(err, tx) {
+        Post.updateAll({content: 'post-a'}, {content: 'post-b'}, {transaction: tx},
+          function(err, changes) {
+            tx.commit(function(err) {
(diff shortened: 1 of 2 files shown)


test/postgresql.transaction.test.js
```diff
@@ -109,6 +109,33 @@ describe('transactions', function() {
     it('should see the committed insert', expectToFindPosts(post, 1));
   });
 
+  describe('on the model', function() {
+    var p1Content = {title: 'p1', content: 'post-a'};
+    var p2Content = {title: 'p2', content: 'post-a'};
+
+    before(function(done) {
+      Post.create(p1Content, function(err, p1) {
+        Post.create(p1Content, function(err, p2) {
```
b-admike: I think you mean `Post.create(p2content, function (err, p2) {` here

jannyHou: sorry it was changed in the old PR, but everything messed up due to the cherry-pick history and I create this one, please hold on review, I need to change something.

loay: As a suggestion, you can also do 

      async.series([    // or parallel depending on the situation 
        function(next) {
          Post.create(content, function(err, post) {
           .....
        },
        function(next) {
          Post.create(content, function(err, post) {
        ...

test/postgresql.transaction.test.js
```diff
@@ -109,6 +109,33 @@ describe('transactions', function() {
...
+          done();
+        });
+      });
+    });
+
+    it('should work when operating directly on the model', function(done) {
+      Post.beginTransaction(Transaction.READ_COMMITTED, function(err, tx) {
+        Post.updateAll({content: 'post-a'}, {content: 'post-b'}, {transaction: tx},
```
b-admike: I didn't know we can pass two filters like this. I thought it would be something like `where: {or: [{ content: 'post-a'}, {content: 'post-b'}]}`. Ignore if this achieves the same result :)

test/postgresql.transaction.test.js
```diff
@@ -109,6 +109,33 @@ describe('transactions', function() {
     it('should see the committed insert', expectToFindPosts(post, 1));
   });
 
+  describe('on the model', function() {
+    var p1Content = {title: 'p1', content: 'post-a'};
+    var p2Content = {title: 'p2', content: 'post-a'};
```
loay: Did u mean `post-b` as per ur test?

test/postgresql.transaction.test.js
jannyHou: we should still be about to pass an array into `Model.create()`, I would prefer to use batchCreate, while anyway, the test seems not able to reproduce the error.

jannyHou:
Closing in favor of https://github.com/strongloop/loopback-connector/pull/109

Issue #126, opened 2016-02-16
Geopoint comparison fails due to non-existent operator


daamsie:
The pgsql connector is using the = operator to compare geopoints when it should be using ~=

I have code like to this

```
var location = new loopback.GeoPoint({lat: ctx.args.data.location.lat, lng: ctx.args.data.location.lng});

Stop.app.models.Place.find(
      { where: { and: [
            {userid: ctx.args.data.userid},
            {location: location}
          ]
        }
      }, ...
```

With debug on, the console returns a query like this. 

```
SELECT "place_id","userid","country_id" FROM "public"."places" 
WHERE ("userid"=$1) AND ("location"= point($2,$3)) ORDER BY "place_id" LIMIT 1

```

Postgresql returns the following error: 
`operator does not exist: point = point`

The solution to this seems to be to adjust the operator to not use = , but use ~= instead.

0candy:
@daamsie Could you please provide a sample sandbox to reproduce your issue? Thanks!
See http://loopback.io/doc/en/contrib/Reporting-issues.html#bug-report

daamsie:
I'm not really using loopback much right now and am short on time so I'm overloaded with other things to do so I'm unlikely to get you this sandbox I'm afraid. However, I think my description above describes the issue clearly enough and even includes the solution to the issue. Hopefully you can act on it.

timlind:
I'm seeing a similar issue except that the generated SQL simply doesn't have an operator at all. This is true for using `{ near: {lat,lng}}` as well.

SQL generated: ... `AND ("geolocation"point($4,$5)) ` ...
SQL error: `error: syntax error at or near "point"`
Query via REST: `{"geolocation":{"lat":36.778261,"lng":-119.41793239999998}}`
Alternative query via REST: `{"geolocation":{"near":{"lat":36.778261,"lng":-119.41793239999998}}}`

loopback@2.36.0
loopback-connector-postgresql@2.7.0
loopback-datasource-juggler@2.53.0

gregdingle:
#191  might be related.

ofekd:
I'm also getting the error @timlind described -- can someone point me to the responsible files? I'll try and fix the issue

gregdingle:
[...]

PR #274, opened 2017-07-13
Fix geo operator

ssh24:
### Description

Currently, postgresql does not support filtering geopoint fields. This is because postgresql uses the base connector level **buildWhere** function to build the query. Not all SQL connectors have the exact way to compare geopoint values and that is why postgresql fails.

This PR overrides two SQL connector functions **buildWhere** and **_buildWhere** functions and adds the additional check

```js
if (p.type.name === 'GeoPoint')
   stmt.merge(columnName + '~=').merge(columnValue);
```

connect to https://github.com/strongloop/loopback-connector-postgresql/issues/126

### DIFF
--- a/test/postgresql.test.js
+++ b/test/postgresql.test.js
@@ -69,6 +69,10 @@ describe('postgresql connector', function() {
     created = new Date();
   });
 
+  after(function(done) {
+    Post.destroyAll(done);
+  });
+
   describe('Explicit datatype', function() {
     before(function(done) {
       db = getDataSource();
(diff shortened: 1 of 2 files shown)


ssh24:
@slnode test please

test/postgresql.test.js
```diff
@@ -279,15 +283,44 @@ describe('postgresql connector', function() {
...
+          done();
+        });
+      });
+    });
+
+    it('should support complicated where filter', function(done) {
+      Post.find({where: {and: [{loc: loc}, {title: {like: '%t%'}}]}}, function(err, post) {
+        should.not.exist(err);
+        should.exist(post);
+        post.length.should.equal(0);
```
b-admike: I suggest the filter is valid and returns a post here

ssh24: That could also be done. But the previous test checks for the case where a `post` is returned so decided to check on this one where `post` is empty.

PR #278, opened 2017-07-18
[WIP-Don't Merge] Set ilike/inlike to true

loay:
### Description
Set ilike and inlike to true. 

### Checklist

### DIFF
--- a/test/init.js
+++ b/test/init.js
@@ -56,8 +56,8 @@ global.getDataSource = global.getSchema = function(useUrl) {
 };
 
 global.connectorCapabilities = {
-  ilike: false,
-  nilike: false,
+  ilike: true,
+  nilike: true,
 };
 
 global.sinon = require('sinon');


ssh24:
@loay I think I know why this is tests are originally skipped. The `ilike` format for memory and mongodb connector does not support pattern matching. Whereas postgres supports pattern matching. 

Meaning if I have a model instance:
```json
{
  "name": "John Lennon"
}
```

Juggler tests run ilike in this format: `User.find({where: {name: {ilike: 'john'}}}, cb)` whereas for postgresql it needs to be `User.find({where: {name: {ilike: 'john%'}}}, cb)`. 

**TLDR:** Memory/mongodb connector returns an instance even if the passed string is a subset of the object found whereas postgres looks for exact matches.

I don't think we need to enable this tests since we have connector specific tests on ilike/nilike for postgres [here](https://github.com/strongloop/loopback-connector-postgresql/blob/21f09b229e458fd4520369de982bf08a69f2e679/test/postgresql.test.js#L383).

loay:
Yeah I agree. I think it is best to keep Postgresql related tests here only.

Issue #217, opened 2017-02-18
Altering table from string to date through autoupdate failing


Strate:
### Bug or feature request

- [x] Bug
- [ ] Feature request

### Description of feature (or steps to reproduce if bug)
I ran into this issue while update from loopback 3.2.0 to loopback 3.3.0.

loopback@3.3.0 introduced change in builtin model Role. It changes type of "created" field from "string" to timestamp. Standard autoupdate function fails to alter db column with error "column "created" cannot be cast automatically to type timestamp with time zone"

mikeodell77:
I am also having this issue.

bajtos:
Hello, thank you for reporting the issue. I am afraid we may not have enough bandwidth to fix it right now. Did you managed to work around this problem by updating the columns from "string" to "timestamp" manually? Would you mind contributing the fix yourself?

ssh24:
@Strate Thank you for reporting the issue. Do you have sample reproduction steps to reproduce this issue?

ssh24:
This issue is fixed and should be available with the next released version of the connector. Thank you for your patience.

PR #279, opened 2017-07-18
Fix alter property in autoupdate

ssh24:
### Description

When using **autoupdate** and altering a column property type, for example from `String` type to `Date`, postgresql requires explicit type casting. This fix allows the connector to handle the type casting when doing such alteration.

connect to https://github.com/strongloop/loopback-connector-postgresql/issues/217

### DIFF
--- a/test/postgresql.autoupdate.test.js
+++ b/test/postgresql.autoupdate.test.js
@@ -5,13 +5,63 @@
 
 'use strict';
 var assert = require('assert');
-var ds;
+var ds, properties, SimpleEmployee;
 
 before(function() {
   ds = getDataSource();
 });
 
-describe('PostgreSQL connector', function() {
+describe('autoupdate', function() {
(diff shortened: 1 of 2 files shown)


test/postgresql.autoupdate.test.js
```diff
@@ -5,13 +5,64 @@
...
+        function(err, props) {
+          assert(!err);
+          assert.equal(props[0].dataType, 'text');
+          assert.equal(props[1].dataType, 'integer');
+          assert.equal(props[2].dataType, 'text');
+          done();
+        });
+    });
+
+    it('perform autoupdate', function(done) {
```
b-admike: I think this can be combined with the next `it()` since you are not doing any assertions here.

ssh24: True. Fixed.

PR #280, opened 2017-07-25
Add CODEOWNER file

dhmlau:
connect to strongloop-internal/scrum-apex#278


slnode:

Issue #282, opened 2017-07-27
double precision type bug


fosk06:
# Description/Steps to reproduce
I have a posteGresql database with a table "coordinate" with a field "lng" wich type is "double precision" .
there the request to create it
CREATE TABLE public.coordinate
(
    id uuid NOT NULL,
    event uuid,
    users uuid,
    address character varying(255) COLLATE pg_catalog."default" NOT NULL,
    lng double precision NOT NULL,
    lat double precision NOT NULL
)
WITH (
    OIDS = FALSE
)
TABLESPACE pg_default;

First  i created the models with a script using datasource.discoverModelDefinitions and
ds.discoverSchemas.
After some bug tracking, i found that the bug come for the model , with the fields "lng" and "lat" :
"lng": {
      "type": "Number",
      "required": true,
      "length": null,
      "precision": 53,
      "scale": null,
      "postgresql": {
        "columnName": "lng",
        "dataType": "double precision",
        "dataLength": null,
        "dataPrecision": 53,
        "dataScale": null,
        "nullable": "NO"
      }
    },
    "lat": {
      "type": "Number",
      "required": true,
      "length": null,
      "precision": 53,
      "scale": null,
      "postgresql": {
        "columnName": "lat",
        "dataType": "double precision",
        "dataLength": null,
        "dataPrecision": 53,
        "dataScale": null,
        "nullable": "NO"
      }
    }
then i'm trying to use the autoUpdate method in another script : 
var ds = loopback.createDataSource('postgresql', require('../datasources').dbpsql)
ds.autoupdate('coordinate', (err,result) => {
console.error(error)
})
I have then an error displayed : 
" error: syntax error at or near "(" "

# Expected result

# Additional information

darwin x64 7.8.0
├── loopback@3.9.0
├── loopback-boot@2.25.0
├── loopback-component-explorer@4.2.0
├── loopback-connector-postgresql@3.0.2

ssh24:
@fosk06 I don't know what is causing this issue on your end. Maybe some syntax error on the script you wrote? 

I have tried on my end and cannot reproduce this.

**model-definition.json**
```json
{
  "name": "City",
  "base": "PersistedModel",
  "idInjection": true,
  "options": {
    "validateUpsert": true
  },
  "properties": {
    "name": {
      "type": "string"
    },
    "lng": {
      "type": "number",
      "postgresql": {
        "dataType": "double precision"
      }
    }, 
    "lat": {
      "type": "number",
[...]

PR #283, opened 2017-07-30
Bug in `double precision/real` data type

ssh24:
### Description
If I have an existing table in PostgreSQL with fields with either of type **double precision** or **real**, then discover the table schema as a loopback model, change the model definition and then perform autoupdate, PostgreSQL will throw an error. 

This is because by default, the **double precision** has a precision set to 53 and **real** has a precision set to 24 in postgres and neither of the data types accept parameterized precision values. Both of **double precision** and **real** are child types of **float** which takes parameterized values to determine the children type. 

i.e `float(n)` if n >=1 and n <=24 it is of type **real** and if n >=25 and n <= 53, it is of type **double precision**. Read more [here](https://www.postgresql.org/docs/9.4/static/datatype-numeric.html)

Since neither **double precision** nor **real** takes a parameterized value for specifying the precision (the precision is automatically set by using **float** data type) but has a default precision value, autoupdate fails. 

This fix overrides the  `discoverModelProperties` from [base connector](https://github.com/strongloop/loopback-connector/blob/master/lib/sql.js#L1877) and changes the data type to **float** if the data type is either **double precision** or **real**.

fixes https://github.com/strongloop/loopback-connector-postgresql/issues/282

### DIFF
--- a/test/postgresql.discover.test.js
+++ b/test/postgresql.discover.test.js
@@ -8,9 +8,10 @@ process.env.NODE_ENV = 'test';
 require('should');
 
 var assert = require('assert');
+var _ = require('lodash');
 
 var DataSource = require('loopback-datasource-juggler').DataSource;
-var db;
+var db, City;
 
 before(function() {
   var config = getDBConfig();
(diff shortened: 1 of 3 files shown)


test/postgresql.discover.test.js
```diff
@@ -105,6 +106,67 @@ describe('Discover models including other users', function() {
...
+      }},
+    });
+    db.automigrate(done);
+  });
+
+  after(function(done) {
+    City.destroyAll(done);
+  });
+
+  it('discover model properties', function(done) {
```
b-admike: Suggestion: `coerces real and dp types to float on model prop discovery`

Issue #266, opened 2017-06-23
Nested object filtering on a JSON field


ssh24:
# Description
Allow nested object filtering on a JSON field to query on such fields. [Docs](http://www.postgresqltutorial.com/postgresql-json/) on how to support JSON in postgresql.

End product should allow the users to filter in this way:
```
Model.find({where: {'contact.email.primary': 'yyy'}}, cb);
```

# Acceptance Criteria
- The following loopback filters should be supported on a postgres JSON field:
  - [ ] where
  - [ ] order
  - [ ] fields
- Robust test cases to make sure they work as expected
- Document an example or two to show how this can be achieved

PR #304, opened 2017-10-27
Add basic json query support

zbarbuto:
### Description

Add basic support for json querying (where and order-by only - no field limiting to keep it simple).

#### Related issues

- #266 (was closed by bot)

### Checklist

Incidentally, I notice this repo doesn't support ES6 syntax online some of the other loopback repos now do 😞

### DIFF
--- a/lib/postgresql.js
+++ b/lib/postgresql.js
@@ -321,6 +321,37 @@ function escapeLiteral(str) {
   return escaped;
 }
 
+/*
+ * Check if a value is attempting to use nested json keys
+ * @param {String} property The property being queried from where clause
+ * @returns {Boolean} True of the property contains dots for nested json
+ */
+function isNested(property) {
+  return property.split('.').length > 1;
+}
+
+/*
+ * Overwrite the loopback-connector column escape
+ * to allow querying nested json keys
+ * @param {String} model The model name
+ * @param {String} property The property name
+ * @returns {String} The escaped column name, or column with nested keys for deep json columns
+ */
+PostgreSQL.prototype.columnEscaped = function(model, property) {
+  if (isNested(property)) {
+    // Convert column to PostgreSQL json style query: "model"->>'val'
+    var self = this;
+    return property
+      .split('.')
+      .map(function(val, idx) { return (idx === 0 ? self.columnEscaped(model, val) : escapeLiteral(val)); })
+      .reduce(function(prev, next, idx, arr) {
+        return idx == 0 ? next : idx < arr.length - 1 ? prev + '->' + next : prev + '->>' + next;
+      });
+  } else {
+    return this.escapeName(this.column(model, property));
+  }
+};
+
 /*!
  * Escape the name for PostgreSQL DB
  * @param {String} name The name
@@ -499,6 +530,12 @@ PostgreSQL.prototype._buildWhere = function(model, where) {
       // The value is not an array, fall back to regular fields
     }
     var p = props[key];
+
+    if (p == null && isNested(key)) {
+      // See if we are querying nested json
+      p = props[key.split('.')[0]];
+    }
+
     if (p == null) {
       // Unknown property, ignore it
       debug('Unknown property %s is skipped for model %s', key, model);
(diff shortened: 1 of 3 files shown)


zbarbuto:
@slnode test please

lib/postgresql.js
```diff
@@ -500,9 +524,14 @@ PostgreSQL.prototype._buildWhere = function(model, where) {
     }
     var p = props[key];
     if (p == null) {
-      // Unknown property, ignore it
-      debug('Unknown property %s is skipped for model %s', key, model);
-      continue;
+      if (isNested(key)) {
```
b-admike: would `p` be null if the property has nested json keys? In other words, could we move this condition outside of the `p == null` check? I'm worried that this is unreachable code, unless I'm missing something.

test/postgresql.test.js
```diff
@@ -658,6 +658,67 @@ describe('postgresql connector', function() {
...
+          postgresql: {
+            dataType: 'json',
+          },
+        },
+      });
+
+      db.automigrate(function(err) {
+        if (err) return done(err);
+        Customer.create([{
+          address: {
```
b-admike: Can we add test/s for 1 more level of nested keys (for checking the recursive calls in https://github.com/strongloop/loopback-connector-postgresql/pull/304/files#diff-ac54005f9bbf753aa45ceebc4a032628R341, or would that be redundant? I.e.
```
Customer.create([{
  name: {
    fullName: {
      firstName: 'foo'
      lastName: 'bar'
    }
  }
}], function (err, customers) {
...
````

File: lib/postgresql.js (hunk 1 above)
Comment (zbarbuto) [reply lib/postgresql.js]: `p` can be null if the property is not defined on the model. Model might have `foo` property but `foo.bar` is being queried so `props[key]` is null but `isNested(key)` is true.

On the flip side if we try to query `baz` (which doesn't exist on the model and isn't a nested query) it should fall through so nothing unreachable.

However you do make a point that `p` could be null if it is nested and the property doesn't exist on the model. Will fix that.

File: test/postgresql.test.js (hunk 2 above)
Comment (zbarbuto) [reply test/postgresql.test.js]: Definitely not redundant since it uncovered a bug 😅 

Syntax for deeply nested needed to be `foo->bar->>baz` since `->>` returns the value at the path as text but `->` returns the actual object at the path.

https://www.postgresql.org/docs/9.5/static/functions-json.html

**Comment by zbarbuto** [comment]
@slnode test please

**Comment by zbarbuto** [comment]
Unable to see failures due to IBM firewall - but tests were passing locally and seem to be failing in other pull requests so possibly unrelated.

**Comment by b-admike** [comment]
@slnode test please

File: test/postgresql.test.js (hunk 2 above)
Comment (b-admike) [reply test/postgresql.test.js]: Ah great! thank you for the link :-).

**Comment by b-admike** [comment]
@slnode test please

**Comment by b-admike** [comment]
It was failing for the following reason:
```
> loopback-connector-postgresql@3.1.0 test /home/jenkins/workspace/nb/loopback-connector-postgresql~master/504599fa
> mocha -R spec --timeout 10000 --require test/init.js test/*.test.js

module.js:471
    throw err;
    ^

Error: Cannot find module 'pseudomap'
    at Function.Module._resolveFilename (module.js:469:15)
    at Function.Module._load (module.js:417:25)
    at Module.require (module.js:497:17)
    at require (internal/module.js:20:19)
    at Object.<anonymous> (/home/jenkins/workspace/nb/loopback-connector-postgresql~master/504599fa/node_modules/lru-cache/index.js:7:11)
    at Module._compile (module.js:570:32)
```
although I don't see that module used here. I've kicked off another build.

zbarbuto:
Correct, that wasn't added as part of this PR.

Apparently there were some issues with the internal registry recently - probably related

https://github.com/strongloop/loopback-datasource-juggler/pull/1513

b-admike:
@zbarbuto all the tests are passing now 🎉. Please squash your commits into one since they're all for one logical change, and we should be able to land.

zbarbuto:
Squished

b-admike:
Thanks @zbarbuto!

PR #317, opened 2018-02-07
Added fix to support regexp for nested query param

MantasR:
### Description
`
GET /path/to/api/model
filter={
    "param.subparam": {
        "regexp": "^keyword"
    }
}
`
Generated SQL params ends up being:
`
[undefined]
`

I have added a regexp parser so generated sql params would be:
`
['^keyword']
`

### Checklist


slnode:

PR #322, opened 2018-02-28
added possibility to add a file path to seek password

gfrancqu:
### Description
Added the ability to add a "passwordFile" field to the datasources.json to allow to store the password into a file.
This can be done programatically but doing so will break the use of the loopback cli utility (the cli utility only pick datasources available in datasources.json)
For example this can be usefull when using docker secret as storing secret in environnement may be judged insecure see moby/moby#30866

### Checklist

- [ x] Code conforms with the [style


slnode:

PR #329, opened 2018-04-24
Stop executing query if transaction has ended

sahilarora946:
### Description
This fixes the issue described in #230 

#### Related issues
#129 

Refer to `rollback` function in https://github.com/zbarbuto/loopback-connector-postgresql/blob/1361460cea86fadd14a68aa8fe9c55f14ecb9a6d/lib/transaction.js#L58-L86

After the `ROLLBACK` is executed, callback will be sent to the event loop to be executed. Event loop can pick any of the available callbacks to execute. It can pick another function that can do database calls in the same transaction and because releaseConnection has not been called, the statement will execute.
To stop the execution,`connection.txId` should be set to null without waiting for the callback of `ROLLBACK` to be called

### DIFF
--- a/lib/transaction.js
+++ b/lib/transaction.js
@@ -69,7 +69,7 @@ function mixinTransaction(PostgreSQL) {
     });
     //If we don't set txId to null and wait for the callback
     //of ROLLBACK query to execute, another query can be executed in the
-    //same transaction because the callback will be called asynchronously 
+    //same transaction because the callback will be called asynchronously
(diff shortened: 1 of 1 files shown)


slnode:

zbarbuto:
Thanks for the PR - could you please fix up the commit message to satisfy the linter.

sahilarora946:
Created another pull request for this. Kindly ignore this one.

PR #330, opened 2018-04-24
stop execution if rollback is called

sahilarora946:
### Description
This fixes the issue described in #230 

#### Related issues
#129

Refer to rollback function in https://github.com/zbarbuto/loopback-connector-postgresql/blob/1361460cea86fadd14a68aa8fe9c55f14ecb9a6d/lib/transaction.js#L58-L86

After the `ROLLBACK` is executed, callback will be sent to the event loop to be executed. Event loop can pick any of the available callbacks to execute. It can pick another function that can do database calls in the same transaction and because releaseConnection has not been called, the statement will execute.
To stop the execution,`connection.txId` should be set to null without waiting for the callback of `ROLLBACK` to be called

### DIFF
--- a/lib/transaction.js
+++ b/lib/transaction.js
@@ -67,6 +67,12 @@ function mixinTransaction(PostgreSQL) {
       self.releaseConnection(connection, err);
       cb(err);
     });
+    //If we don't set txId to null and wait for the callback
+    //of ROLLBACK query to execute, another query can be executed in the
+    //same transaction because the callback will be called asynchronously
+    if (typeof connection.autorelease === 'function') {
+      connection.txId = null;
+    }
   };
 
   PostgreSQL.prototype.releaseConnection = function(connection, err) {


slnode:

zbarbuto:
@slnode test please

sahilarora946:
@zbarbuto why are these tests failing?

dhmlau:
@slnode test please

dhmlau:
@zbarbuto , sorry about the late reply.  There was an CI issue that we ran into before.  I thought it's been fixed. Let me kick off CI again to check.

dhmlau:
CI all passed. 
@sahilarora946 , could you please rebase your PR?  Thanks.

sahilarora946:
@dhmlau rebase done

dhmlau:
@slnode  test please

dhmlau:
@zbarbuto @strongloop/sq-lb-apex , PTAL.  
I'm not very familiar with transaction.

zbarbuto:
LGTM - I'm a bit out of the loop on it also, though. @raymondfeng comes to mind, maybe give it a quick once over and then should be able to merge?

b-admike:
Thank you for your contribution @sahilarora946. Taking a look. I'm wondering if we can simulate the behaviour that causes this issue in a test.

b-admike:
the PR LGTM, but I recommend adding a test which can simulate the scenario/bug we are fixing to avoid regressions in the future. @raymondfeng can you PTAL?

dhmlau:
Released as 3.3.2.

PR #331, opened 2018-04-24
update CODEOWNERS

dhmlau:
### Description
Add our previous maintainers to the "alumni members" section.

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -3,4 +3,7 @@
 # the last matching pattern has the most precendence.
 
 # Core team members from IBM
-* @kjdelisle @jannyHou @loay @b-admike @ssh24 @virkt25 @dhmlau @zbarbuto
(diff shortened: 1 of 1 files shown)


CODEOWNERS
```diff
@@ -3,4 +3,7 @@
 # the last matching pattern has the most precendence.
 
 # Core team members from IBM
-* @kjdelisle @jannyHou @loay @b-admike @ssh24 @virkt25 @dhmlau @zbarbuto
+* @jannyHou @b-admike @virkt25 @dhmlau @zbarbuto
+
+# Alumni members
+* @kjdelisle @loay @ssh24
```
virkt25: I think this line needs to be commented out so they aren't treated as active members by GitHub.

dhmlau: you're right. otherwise the last line will take precedence.  thanks!

PR #318, opened 2018-02-09
add support for UPDATE RETURNING

timgit:
### Support UPDATE ... RETURNING

Adds `rows` from pg driver after an UPDATE if RETURNING specified

### Checklist

- [x] New tests added or existing tests modified to cover all changes
- [x] Code conforms with the [style guide](http://loopback.io/doc/en/contrib/style-guide.html)

PR #334, opened 2018-06-26
Adds support for INSERT RETURNING

timgit:
### Description

Adds support to fetch rows from INSERT with a RETURNING clause.  Since the behavior is the same as UPDATE RETURNING, I wanted to see if I could get away with 

#### Related issues

Identical in behavior to PR #318.

### Checklist


slnode:

PR #341, opened 2018-08-29
Export table comments in description field

MelleB:
### Description
This patch exports comments on columns as descriptions in the models. To be used for spec generation downstream.

#### Related issues
- none

### Checklist

### DIFF
--- a/test/postgresql.discover.test.js
+++ b/test/postgresql.discover.test.js
@@ -339,4 +339,14 @@ describe('Discover and build models', function() {
       });
     });
   });
+
+  it('should build a model from discovery with comments', function(done) {
+    db.discoverAndBuildModels('TestGeo', {schema: 'public'}, function(err, schema) {
+      var prop = schema.Testgeo.definition.properties.loc.postgresql;
+      assert(prop);
+      assert.equal(prop.description, 'The location of property of TestGeo');
+      done();
+    });
+  });
 });
+
--- a/lib/discovery.js
+++ b/lib/discovery.js
@@ -149,6 +149,11 @@ function mixinDiscovery(PostgreSQL) {
             || r.dataType === 'smallint')
             r.dataPrecision = null;
           r.type = self.buildPropertyType(r, options);
+
+          if (r.columnComment) {
+            r.postgresql = {description: r.columnComment};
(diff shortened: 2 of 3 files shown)


slnode:

dhmlau:
@slnode test please

test/postgresql.discover.test.js
```diff
@@ -339,4 +339,14 @@ describe('Discover and build models', function() {
       });
     });
   });
+
+  it('should build a model from discovery with comments', function(done) {
+    db.discoverAndBuildModels('TestGeo', {schema: 'public'}, function(err, schema) {
+      var prop = schema.Testgeo.definition.properties.loc.postgresql;
+      assert(prop);
+      assert.equal(prop.description, 'The location of property of TestGeo');
```
jannyHou: It would be good if you can assert [the comment for table](https://github.com/strongloop/loopback-connector-postgresql/pull/341/files#diff-1d86d0e19a7dc8c0ecc368e25741e1a0R59) as well.

test/postgresql.discover.test.js
```diff
@@ -339,4 +339,14 @@ describe('Discover and build models', function() {
       });
     });
   });
+
+  it('should build a model from discovery with comments', function(done) {
+    db.discoverAndBuildModels('TestGeo', {schema: 'public'}, function(err, schema) {
+      var prop = schema.Testgeo.definition.properties.loc.postgresql;
```
bajtos: Is there any particular reason for scoping the description to `postgresql`? I would personally prefer the description to be LoopBack-wide.

```ts
const prop = schema.Testgeo.definition.properties.loc;
assert.equal(prop.description, 'The location of property of TestGeo');
```

When the description is set directly on the property, then tools like loopback-swagger (powering loopback-component-explorer in LB v3) and `@loopback/repository-json-schema` (in LB v4) will pick it up and use it to describe the property in Swagger/OpenAPI spec.

bajtos:
This is out of scope of this pull request, but would it make sense to support property descriptions in automigration too? I.e. `lb4 automigrate` should use model & property `description` to set PostgreSQL's comment field on a table or a column.

PR #345, opened 2018-10-18
Type 'double precision' mapped to String

ataft:
On line 144, dataType "real" and "double precision" are changed to "float", but buildPropertyType has no case for "FLOAT":
https://github.com/strongloop/loopback-connector-postgresql/blob/master/lib/discovery.js#L144

This change simply adds a case statement for 'FLOAT'

### DIFF
--- a/lib/discovery.js
+++ b/lib/discovery.js
@@ -342,6 +342,7 @@ function mixinDiscovery(PostgreSQL) {
       case 'NUMERIC':
       case 'REAL':
       case 'DOUBLE PRECISION':
+      case 'FLOAT':
       case 'SERIAL':
       case 'BIGSERIAL':
         return 'Number';


ataft:
@dhmlau I'm not so great with git and don't know how to change the commit message.

With regards to the tests, also not my strong suit, but the current "postgresql.discover.test.js" should cover this the 'double precision' type:
https://github.com/strongloop/loopback-connector-postgresql/blob/master/test/postgresql.discover.test.js#L113

dhmlau:
@ataft, see if you can follow this instruction to change the commit message https://help.github.com/articles/changing-a-commit-message/.  If not, don't worry, I can modify it for you.

Thanks for pointing out the existing test.

dhmlau:
Let's get the commit message fixed so that CI can run successfully first.  Thanks!

dhmlau:
@slnode test please

dhmlau:
Thanks @ataft.  Your PR has landed.  Thanks for the contribution!

PR #351, opened 2018-11-21
[DISCUSSION] Deep JSON key fields filter

u8sand:
### Description
A working hack of this feature to illuminate what needs to be done. I'm open to suggestions for things I should improve but wanted to put this here for anyone interested in the feature. This is not meant to be merged in its current state.

#### Discussion
Supporting filtering by json fields should be possible with the postgresql connector. That-is, a find that looks like so:
```json
{
  "where": {
    "address.city": "Hill Valley",
  },
  "fields": ["address.street.number"],
}
```

Should result in
```json
{
  "address": {
    "street": {
      "number": 56
    }
  }
}
```

A number of issues make this difficult, not least of all because strongloop/loopback-datasource-juggler pre-filters unrecognizable fields in the query. A workaround is to alter the `strict` setting on the model definition; see https://github.com/strongloop/loopback-connector-postgresql/blob/e58ffccc58232d50e01620aa54951f398754c317/test/postgresql.test.js#L692-L702 for full discussion.

Once we get the fields passed the juggler, we can deal with the filters in the `buildColumnNames` override. The same code used in #304 to create the where clause can be used for the select clause, but unfortunately the postgresql output becomes no longer identifiable.
```
 ?column? 
----------
 56
```

To mitigate this, we can use an `as` on nested fields, and because PostgreSQL doesn't like `.` for names, we can replace those with `__` among other things.
```
 address__street__number 
-------------------------
 56
```

To capture this we need to override `fromRow`; and construct a deep dictionary using the key/value.
```
{
  "address": {
    "street": {
      "number": 56
    }
  }
}
```

Finally, the juggler tries to do our job once again--unfortunately this time, there doesn't seem to be a fix beyond patching upstream. It filters the fields again after they've been assembled, this removes the entire `address` since it doesn't appear in `fields`. Just adding it wouldn't be good else we'd get the entire deep object... I've patched it upstream here: https://github.com/u8sand/loopback-datasource-juggler/commit/81ed9dfccfb21537043634daa8f19eb78a97e639

With the patch applied, it works, except that we can't identify sub-field types and end up with a string instead of an integer.

**I hope we can somehow mitigate some duplication happening in the `loopback-datasource-juggler` before this becomes a viable feature.**

I also need to add more tests, deal with the other filter formatting, and carefully review things to ensure we're not introducing any vulnerabilities (more field escaping may be in order).

Any suggestions welcome to make this feature more of a reality.

#### Related issues

- Finishes off what was started by #304 by adding fields
- #266
- ... definitely others

### Checklist

- [ ] Received feedback
- [ ] Patch strongloop/loopback-datasource-juggler
- [x] Code conforms with the [style guide](http://loopback.io/doc/en/contrib/style-guide.html)

### DIFF
--- a/lib/postgresql.js
+++ b/lib/postgresql.js
@@ -413,6 +413,112 @@ PostgreSQL.prototype.applyPagination = function(model, stmt, filter) {
   return stmt.merge(limitClause);
 };
 
+/**
+ * Assign a value to a nested dictionary, with a list of keys.
+ * @param {object} obj The target of the assignment
+ * @param {array} keys A list of keys to use for assignment
+ * @param {any} value The value to assign the last key to.
+ * @returns {object} Model data object
+ */
+function deepAssign(obj, keys, value) {
+  if(keys.length > 0) {
+    obj[keys[0]] = deepAssign(obj[keys[0]] || {}, keys.slice(1), value)
+    return obj
+  } else {
+    return value
+  }
+}
+
+/**
+ * Transform the row data into a model data object
+ * @param {string} model Model name
+ * @param {object} rowData An object representing the row data from DB
+ * @returns {object} Model data object
+ */
+PostgreSQL.prototype.fromRow = PostgreSQL.prototype.fromDatabase =
+function(model, rowData) {
+  if (rowData == null) {
+    return rowData;
+  }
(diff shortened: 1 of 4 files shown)


slnode:

Issue #364, opened 2019-02-19
Connector fails during initialization - Invalid property path for getNestedPropertyDefinition


SippieCup:
It seems the newest pull request in loopback-connector (https://github.com/strongloop/loopback-connector/pull/136) breaks the postgreql connector.

# Description/Steps to reproduce

npm install, then run any psql connected application. 

# Expected result
```
NODE_ENV=production /opt/elasticbeanstalk/node-install/node-v10.15.0-linux-x64/bin/node .
/var/app/current/node_modules/loopback-connector/lib/connector.js:159
    throw new Error(g.f('Invalid property path'));
    ^

Error: Invalid property path
    at Function.Connector.getNestedPropertyDefinition (/var/app/current/node_modules/loopback-connector/lib/connector.js:159:11)
    at PostgreSQL.Connector.getPropertyDefinition (/var/app/current/node_modules/loopback-connector/lib/connector.js:139:20)
    at PostgreSQL.SQLConnector.column (/var/app/current/node_modules/loopback-connector/lib/sql.js:376:19)
    at /var/app/current/node_modules/loopback-connector/lib/sql.js:157:27
    at Array.forEach (<anonymous>)
    at PostgreSQL.SQLConnector.searchForPropertyInActual (/var/app/current/node_modules/loopback-connector/lib/sql.js:156:16)
    at /var/app/current/node_modules/loopback-connector-postgresql/lib/migration.js:135:24
    at Array.forEach (<anonymous>)
    at PostgreSQL.getColumnsToAdd (/var/app/current/node_modules/loopback-connector-postgresql/lib/migration.js:133:15)
    at PostgreSQL.getAddModifyColumns (/var/app/current/node_modules/loopback-connector-postgresql/lib/migration.js:117:27)
    at applyPending (/var/app/current/node_modules/loopback-connector-postgresql/lib/migration.js:86:45)
    at /var/app/current/node_modules/loopback-connector-postgresql/lib/migration.js:104:9
    at /var/app/current/node_modules/loopback-connector-postgresql/node_modules/async/lib/async.js:607:21
    at /var/app/current/node_modules/loopback-connector-postgresql/node_modules/async/lib/async.js:246:17
    at iterate (/var/app/current/node_modules/loopback-connector-postgresql/node_modules/async/lib/async.js:146:13)
    at async.eachSeries (/var/app/current/node_modules/loopback-connector-postgresql/node_modules/async/lib/async.js:162:9)
[...]

PR #365, opened 2019-02-19
fix: use original prop name to search fields

b-admike:
### Description
When calling `searchForPropertyInActual`, we should provide the actual prop name since the function itself in [loopback-connector](https://github.com/strongloop/loopback-connector/blob/master/lib/sql.js#L152-L162) calls `self.column` in order to find a column inside the fields which matches the property name.

Discovered this bug when I was working on #364.

### DIFF
--- a/lib/migration.js
+++ b/lib/migration.js
@@ -133,7 +133,7 @@ function mixinMigration(PostgreSQL) {
     propNames.forEach(function(propName) {
       if (self.id(model, propName)) return;
       var found = self.searchForPropertyInActual(
-        model, self.column(model, propName), actualFields);
+        model, propName, actualFields);
       if (!found && self.propertyHasNotBeenDeleted(model, propName)) {
         sql.push('ADD COLUMN ' + self.addPropertyToActual(model, propName));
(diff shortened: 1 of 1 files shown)


elv1s:
thanks for the quick fix! 

Is this a breaking change or a bug caused (exposed?) by the connector change?

b-admike:
> Is this a breaking change or a bug caused (exposed?) by the connector change?
@elv1s It is a bug which is exposed by the connector change and if we land this, then we don't really need https://github.com/strongloop/loopback-connector/pull/138 i.e. all the tests will pass on postgres connector. I'm not sure if the patch in loopback-connector is correct, but for this PR, I'd need to write a test to verify the changes.

dhmlau:
@b-admike , seems like you got approval from @jannyHou.  Anything else we need to do to land this besides rebasing the PR again? Thanks.

b-admike:
@dhmlau I need to write a test(s) for this PR before landing to ensure there won't be regressions. I was having difficulty to write one, but Janny has given me some ideas.

PR #369, opened 2019-03-07
add @elv1s as CODEOWNERS

dhmlau:
### Description
Welcome @elv1s to be the maintainer of the PostgreSQL connector. 🎉 

### Checklist

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -2,8 +2,8 @@
 # Each line is a file pattern followed by one or more owners,
 # the last matching pattern has the most precedence.
 
-# Core team members from IBM
-* @jannyHou @b-admike @dhmlau @zbarbuto
-
 # Alumni members
 # @kjdelisle @loay @ssh24 @virkt25
+
+# Core team members
+* @jannyHou @b-admike @dhmlau @zbarbuto @elv1s


CODEOWNERS
```diff
@@ -2,8 +2,8 @@
 # Each line is a file pattern followed by one or more owners,
 # the last matching pattern has the most precedence.
 
-# Core team members from IBM
-* @jannyHou @b-admike @dhmlau @zbarbuto
-
 # Alumni members
 # @kjdelisle @loay @ssh24 @virkt25
+
+# Core team members from IBM
```
bajtos: `elv1s` is not from IBM, is he?

```suggestion
# Core team members
```

Is there any particular reasons why you are moving the list of active maintainers at the bottom, after alumni members? I believe our other repos use the convention where active members are listed first.

bajtos:
Welcome @elv1s on board! 🎉

CODEOWNERS
dhmlau: You're right. I'll remove the `from IBM`. 

For the order, when i made the changes, I put the active ones at the bottom, seems easier to see.  But I'm good either way, just want to be consistent.  There are other connector repos that should be updated as well. :) 

Active maintainers at the bottom, using `#` as prefix:
https://github.com/strongloop/loopback-connector-mssql/blob/master/CODEOWNERS
https://github.com/strongloop/loopback-connector/blob/master/CODEOWNERS

Alumni at the bottom and using `_` as prefix:
https://github.com/strongloop/loopback/blob/master/CODEOWNERS

bajtos: I can see arguments for both (active at the bottom, alumni at the bottom). No need to bike shed, feel free to pick the order that makes most sense to you.

Out of scope of this PR: should we start building list of alumni maintainers in https://github.com/strongloop/loopback-next/blob/master/CODEOWNERS too?

PR #376, opened 2019-04-30
Run shared tests from both v3 and v4 of juggler

bajtos:
At the moment, our connectors are installing juggler 3.x to run the test suite. When we make a change to juggler@3, cis-jenkins detects downstream dependency and triggers CI build of connectors with the modified juggler version. When we make a change to juggler's master, no downstream builds are triggered.

In the past, we were maintaining multiple branches to test connectors against different juggler versions. E.g. `loopback-2.x` containing code from "master" but installing juggler 2.x for testing. This is rather impractical, because we don't have bandwidth to update those other branches with changes made to master.

In this pull request, I am reworking connector test setup to execute tests from both juggler versions 3.x and 4.x.

### Checklist

### DIFF
--- a/deps/juggler-v4/test.js
+++ b/deps/juggler-v4/test.js
@@ -1,0 +1,37 @@
+// Copyright IBM Corp. 2013,2016. All Rights Reserved.
+// Node module: loopback-connector-mysql
+// This file is licensed under the MIT License.
+// License text available at https://opensource.org/licenses/MIT
+
+'use strict';
+
+// Skip the tests on Node.js versions not supported by juggler v4
+// TODO(bajtos): remove this check when we drop Node.js 6 from our CI matrix
+var nodeMajor = +process.versions.node.split('.')[0];
+if (nodeMajor < 8) {
+  return;
+}
+
+var should = require('../../test/init');
+var juggler = require('loopback-datasource-juggler');
+var name = require('./package.json').name;
+
+describe(name, function() {
+  before(function() {
+    return global.resetDataSourceClass(juggler.DataSource);
+  });
+
+  after(function() {
+    return global.resetDataSourceClass();
+  });
+
+  require('loopback-datasource-juggler/test/common.batch.js');
+  require('loopback-datasource-juggler/test/include.test.js');
+
+  /* TODO: run persistence-hooks test suite too
+  var testHooks = require('loopback-datasource-juggler/test/persistence-hooks.suite.js');
+  testHooks(global.getDataSource(), should, {
+    replaceOrCreateReportsNewInstance: false,
+  });
+  */
+});
--- a/test/init.js
+++ b/test/init.js
@@ -4,7 +4,8 @@
 // License text available at https://opensource.org/licenses/Artistic-2.0
 
 'use strict';
-var DataSource = require('loopback-datasource-juggler').DataSource;
+var juggler = require('loopback-datasource-juggler');
+var DataSource = juggler.DataSource;
 
 var config = require('rc')('loopback', {test: {postgresql: {}}}).test.postgresql;
 
@@ -53,6 +54,13 @@ global.getDataSource = global.getSchema = function(useUrl) {
     // console.log(a);
   };
   return db;
+};
+
+global.resetDataSourceClass = function(ctor) {
+  DataSource = ctor || juggler.DataSource;
+  var promise = db ? db.disconnect() : Promise.resolve();
+  db = undefined;
+  return promise;
 };
 
 global.connectorCapabilities = {
--- a/test/mocha.opts
+++ b/test/mocha.opts
@@ -1,0 +1,4 @@
+--reporter spec
+--timeout 10000
+--require test/init.js
+--exit
--- a/test/postgresql.autoupdate.test.js
+++ b/test/postgresql.autoupdate.test.js
@@ -629,19 +629,27 @@ describe('autoupdate', function() {
       // Table create order is important. Referenced tables must exist before creating a reference.
       // do initial update/creation of referenced tables
       ds.autoupdate(function(err) {
-        assert(!err, err);
+        if (err) {
+          err.message += ' (while running initial autoupdate)';
+          return done(err);
+        }
 
         // do initial update/creation of table with fk
         ds.createModel(orderTest_schema_v1.name, orderTest_schema_v1.properties, orderTest_schema_v1.options);
         ds.autoupdate(function(err) {
-          assert(!err, err);
+          if (err) {
+            err.message += ' (while updating OrderTest schema v1)';
+            return done(err);
+          }
           ds.discoverModelProperties('order_test', function(err, props) {
+            if (err) return done(err);
             // validate that we have the correct number of properties
             assert.equal(props.length, 3);
 
             // get the foreign keys for order_test
             ds.connector.discoverForeignKeys('order_test', {}, function(err, foreignKeys) {
-              assert(!err, err);
+              if (err) return done(err);
+
               // validate that the foreign key exists and points to the right column
               assert(foreignKeys);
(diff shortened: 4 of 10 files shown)


bajtos:
Please note the test failures are coming from juggler v4, we haven't been running those tests yet. I'll fix them before landing this PR if we decide to continue in this direction.

test/mocha.opts
```diff
@@ -0,0 +1,4 @@
+--reporter spec
```
b-admike: 👍

test/init.js
```diff
@@ -4,7 +4,8 @@
 // License text available at https://opensource.org/licenses/Artistic-2.0
 
 'use strict';
-var DataSource = require('loopback-datasource-juggler').DataSource;
+var juggler = require('loopback-datasource-juggler');
```
b-admike: Does this get the root level juggler dev dependency? Would `npm` dedupe cause any issues?

deps/juggler-v4/test.js
```diff
@@ -0,0 +1,30 @@
...
+
+  after(function() {
+    return global.resetDataSourceClass();
+  });
+
+  require('loopback-datasource-juggler/test/common.batch.js');
+  require('loopback-datasource-juggler/test/include.test.js');
+
+  /* TODO: run persistence-hooks test suite too
+  var testHooks = require('loopback-datasource-juggler/test/persistence-hooks.suite.js');
```
b-admike: Looks like this wasn't enabled before - but good to have a place holder and get more test coverage 👏

test/init.js
```diff
@@ -55,6 +56,13 @@ global.getDataSource = global.getSchema = function(useUrl) {
   return db;
 };
 
+global.resetDataSourceClass = function(ctor) {
+  DataSource = ctor || juggler.DataSource;
+  var promise = db ? db.disconnect() : Promise.resolve();
```
b-admike: Have you thought about returning `Promise.reject` with a descriptive message if `db` is not truthy?

test/init.js
bajtos: Yes, this is getting root-level juggler dev dependency. `npm` dedupe is not causing any issues as far as I was able to tell. Here is the `node_modules` layout:

```
deps
  - juggler-v3
    - node_modules
      - async (v2.6.2)
      - loopback-datasource-juggler (v3.30.0)
  - juggler-v4
    - node_modules
      - async (v2.6.2)
      - debug (v4.1.1)
      - depd (v2.0.0)
      - loopback-datasource-juggler (v4.6.3)
node_modules
  - loopback-datasource-juggler (v4.6.3)
  - juggler-v3 -> ../deps/juggler-v3
  - juggler-v4 -> ../deps/juggler-v4
  - (plus a bunch of other packages)
```

It also means that we have three kinds of tests now:

 - The test suite from juggler@3 using juggler@3 runtime
 - The test suite from juggler@4 using juggler@4 runtime
 - The connector-specific test suite in `test/*.js` always using juggler@4 runtime from top-level `node_modules`.

test/init.js
bajtos: It's a valid situation that `db` is not defined. This typically happens when `resetDataSourceClass` is called for the first time, for example when you run `mocha deps/juggler-v3/test.js`. 

Here is the logic of this function:
 - If there is an existing `db` instance left by previous tests, then dispose it (call `disconnect`). We don't want to have a growing number of datasource instances, because each datasource is usually maintaining one or more open connections to the database.
 - (Re)set the datasource class to the new value.

test/init.js
bajtos: After I removed `node_modules` folder and reinstalled dependencies, I got slightly different layout which is closer to what I would expect to see.

```
deps
  - juggler-v3
    - node_modules
      - async (v2.6.2)
      - loopback-datasource-juggler (v3.30.0)
  - juggler-v4 
    # no node_modules created!
node_modules
  - loopback-datasource-juggler (v4.6.3)
  - juggler-v3 -> ../deps/juggler-v3
  - juggler-v4 -> ../deps/juggler-v4
  - (plus a bunch of other packages)

**Comment by bajtos** [comment]
> Please note the test failures are coming from juggler v4, we haven't been running those tests yet. I'll fix them before landing this PR if we decide to continue in this direction.

Should be fixed now, via https://github.com/strongloop/loopback-datasource-juggler/pull/1733  and `loopback-datasource-juggler@4.6.4`

**Comment by bajtos** [comment]
> loopback-connector-postgresql/node=6.x,os=windows — Failed! (795e064)

This test sometimes passes and other times it fails and the error message is not clear about what went wrong :(

```
115 passing (6s)
  1 failing

D:\t\t4938861194367131829dirbfr\node_modules\build-framework\node_modules\mocha-jenkins-reporter\lib\jenkins.js:259
    var n = ++currentSuite.failures;
                          ^

TypeError: Cannot read property 'failures' of null
    at Runner.<anonymous> (D:\t\t4938861194367131829dirbfr\node_modules\build-framework\node_modules\mocha-jenkins-reporter\lib\jenkins.js:259:27)
    at emitTwo (events.js:111:20)
    at Runner.emit (events.js:191:7)
    at Runner.fail (D:\workspace\loopback-connector-postgresql\d60c4a15\node_modules\mocha\lib\runner.js:251:8)
    at Runner.uncaught (D:\workspace\loopback-connector-postgresql\d60c4a15\node_modules\mocha\lib\runner.js:757:8)
    at process.uncaught (D:\workspace\loopback-connector-postgresql\d60c4a15\node_modules\mocha\lib\runner.js:839:10)
    at emitOne (events.js:96:13)
    at process.emit (events.js:188:7)
    at process._fatalException (bootstrap_node.js:300:26)
```

The CI re-runs the tests with DEBUG logs enabled and then the test suite passes. I am proposing to ignore this problem for now. Node.js 6 has reached EOL few days ago, I think it's time to remove it from our CI platforms.

> [cis-jenkins] x64 && linux && nvm && dbs,8 — Failed! (795e064)

I suspect this is intermittent too, I've seen this error few times on my machine too, but not always.

```
09:20:45   1) autoupdate
09:20:45        foreign key constraint
09:20:45          should create, update, and delete foreign keys:
09:20:45 
09:20:45       Uncaught AssertionError [ERR_ASSERTION]: error: duplicate key value violates unique constraint "pg_namespace_nspname_index"
09:20:45       + expected - actual
09:20:45 
09:20:45       -false
09:20:45       +true
09:20:45       
09:20:45       at /home/jenkins/workspace/nb/loopback-connector-postgresql~master/3ed5ebba/test/postgresql.autoupdate.test.js:632:9
```

**Comment by bajtos** [comment]
@slnode test please

**Comment by bajtos** [comment]
All checks have passed this time. This pull request is ready for final review & landing.

File: test/init.js (hunk 2 above)
Comment (b-admike) [reply test/init.js]: Thanks for checking and clarifying @bajtos, that looks good.

File: test/postgresql.autoupdate.test.js [hunk 5]
```diff
@@ -629,19 +629,27 @@ describe('autoupdate', function() {
       // Table create order is important. Referenced tables must exist before creating a reference.
       // do initial update/creation of referenced tables
       ds.autoupdate(function(err) {
-        assert(!err, err);
+        if (err) {
+          err.message += ' (while running initial autoupdate)';
```
b-admike: Kudos for adding these descriptive error messages 👍

deps/juggler-v4/test.js
```diff
@@ -5,6 +5,12 @@
 
 'use strict';
 
+var nodeMajor = +process.versions.node.split('.')[0];
```
b-admike: Is this a temporary solution until we drop Node.js 6.x from our CI matrix in juggler and/or connectors? If so, let's put a comment to remind ourselves down the road.

PR #380, opened 2019-05-09
Drop support for Node.js 6

bajtos:
- Node.js 6.x has reached EOL
- cis-jenkins has a bug where it runs downstream tests of juggler@4 (supporting Node.js 8+ only) on the minimum Node.js version supported by the connector (Node.js 6), which triggers the check skipping juggler@4 tests.

This change will be published as semver-minor release.

### DIFF
--- a/deps/juggler-v4/test.js
+++ b/deps/juggler-v4/test.js
@@ -5,13 +5,6 @@
 
 'use strict';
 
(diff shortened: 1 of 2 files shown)


b-admike:
LGTM 👏🏾

PR #390, opened 2019-08-20
chore: add package-lock.json

aharbis:
### Description

Commits the generated `package-lock.json` file.

#### Related issues

- connect to `none`

### Checklist

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,0 +1,2471 @@
+{
+  "name": "loopback-connector-postgresql",
+  "version": "3.7.0",
+  "lockfileVersion": 1,
+  "requires": true,
+  "dependencies": {
+    "@babel/code-frame": {
+      "version": "7.5.5",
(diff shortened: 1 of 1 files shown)


dhmlau:
I think there was some discussion about _not_ having the package.json in the connector repos.  @bajtos, could you please confirm? 

If so, I think the change should be to add a `.npmrc` file setting the package-lock to false. See https://github.com/strongloop/strong-remoting/blob/master/.npmrc

aharbis:
That could make sense. I had checked issue and code history for the repo and did not see discussion for this, so went ahead and put the PR up.

If the decision is to not commit, I’m happy to submit the PR to add the `.npmrc` file.

dhmlau:
Confirmed with @jannyHou, we don't want `package-lock.json` to be created. So if you could make the change to include the `.npmrc` file, that would be great. Thanks @aharbis!!

aharbis:
Opened PR #391 to add `.npmrc`. Closing this PR.

PR #391, opened 2019-08-20
chore: set package-lock=false in .npmrc

aharbis:
### Description

Confirmed in discussion via PR #390 that we do not want to commit `package-lock.json`. Thus, we should set `package-lock=false` in `.npmrc`.

#### Related issues

- connect to `none`

### Checklist

### DIFF
--- a/.npmrc
+++ b/.npmrc
@@ -1,0 +1,1 @@
+package-lock=false


dhmlau:
@aharbis, thanks for your contribution.  Your PR has landed! 🎉

PR #392, opened 2019-08-24
BIGSERIAL type support 

loryk:
### Description

Add support to  create fields with BIGSERIAL datatype.
This type will be added for options:
- genrated = true
- dataType = BIGINT/BIGSERIAL/BIG...

### DIFF
--- a/lib/migration.js
+++ b/lib/migration.js
@@ -295,7 +295,18 @@ function mixinMigration(PostgreSQL) {
     const self = this;
     const modelDef = this.getModelDefinition(model);
     const prop = modelDef.properties[propName];
+
+    // Read column type property
+    const columnMetadata = this.columnMetadata(model, propName);
+    let colType = columnMetadata && columnMetadata.dataType;
+    if (colType) {
+      colType = colType.toUpperCase();
+    }
+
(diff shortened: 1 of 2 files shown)


slnode:

dhmlau:
@slnode test please

lib/migration.js
```diff
@@ -295,7 +295,18 @@ function mixinMigration(PostgreSQL) {
...
+    // Read column type property
+    const columnMetadata = this.columnMetadata(model, propName);
+    let colType = columnMetadata && columnMetadata.dataType;
+    if (colType) {
+      colType = colType.toUpperCase();
+    }
+
     if (prop.generated) {
+      // For types with prefix BIG should be used BIGSERIAL
+      if (colType && colType.startsWith('BIG'))
```
bajtos: I am worried about a possible confusion this can create for our users, when the framework converts a dataType value `BIGINT` to `BIGSERIAL`.

Why would the developer specify a `BIGINT` data-type when they want to use `BIGSERIAL` instead?

Have you considered to use the `colType` as provided by the user instead of converting all `BIG...` types to `BIGSERIAL`?

Let's discuss.

PR #401, opened 2019-11-24
Use db type jsonb to support dot notation filters

ewrayjohnson:
Using dot notation in query filters does not work because the database column type for object and embedded models is string.  This fixes the problem by changed the type to jsonb.

### DIFF
--- a/lib/migration.js
+++ b/lib/migration.js
@@ -21,7 +21,7 @@ function mixinMigration(PostgreSQL) {
     const tableName = self.table(modelName);
     let _foreignKeys = m.settings.foreignKeys;
     Object.defineProperty(m.settings, 'foreignKeys', {
-      get : function () {
+      get : () => {
         function addForeignKey(keyFrom, modelTo, keyTo) {
           const foreignColumn = self.column(modelName, keyFrom);
           const constraintName = tableName + '_' + foreignColumn + '_fkey';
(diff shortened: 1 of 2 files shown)


slnode:

dhmlau:
@ewrayjohnson, thanks for your PR. Could you please add some tests to verify your code, and also rebase your branch? Thanks.

lib/migration.js
```diff
@@ -479,9 +479,9 @@ function mixinMigration(PostgreSQL) {
     const p = propertyDefinition;
     switch (p.type.name) {
       default:
-      case 'String':
       case 'JSON':
-        return 'TEXT';
+        return 'JSONB';
```
bajtos: What is the impact of this change on existing application? Have you considered introducing a new `p.type.name` called `JSONB`, that will allow app developers using `JSON` to preserve the current behavior, and let them opt into the new JSONB-based storage?

lib/migration.js
```diff
@@ -783,8 +783,7 @@ function mixinMigration(PostgreSQL) {
           if (identical && siKeys.length > 1) {
             if (siKeys.length !== i.keys.length) {
               // lengths differ, obviously non-matching
-              // ??? orderMatched is not defined and not used anywhere else
-              // orderMatched = false;
+              identical = false;
```
bajtos: This change looks unrelated to the other change you are making above. What scenario are you trying to fix with this change?

ewrayjohnson:
I am working on creating tests to show the bug and confirm the fix.

PR #404, opened 2019-12-10
fix: allow string type attribute to be auto-generated in Postgres

agnes512:
fixes https://github.com/strongloop/loopback-next/issues/2398

### _please review it with whitespace hided_

I went through several related issues:
- [postgres canoot generate str type id automatically #811](https://github.com/strongloop/loopback.io/issues/811) 
- [Preserve custom type of auto-generated id property #3602](https://github.com/strongloop/loopback-next/issues/3602)
- [lb3 - Relations on uuid
](https://github.com/strongloop/loopback/issues/4287)

strongloop/loopback-next#2398 is partially fixed in https://github.com/strongloop/loopback-datasource-juggler/pull/1783. 
```
@model()
export class Customer extends Entity {
  @property({
    type: 'string',
    id: true,
    useDefaultIdType: false,
    generated: true,
  })
  custid: string;
``` 

### Issue
The flag `useDefaultIdType: false` is supposed to allow to use custom type of auto-generated id property. But it only allows the default type, and it won't generated the custom id type itself as different DBs have different syntax/setup.

For example, Postgres and MySQL default identity type is INT. And they can be auto-generated by `SERIAL` and `AUTO_INCREAMENT` respectively. 

The model definition above won't work on Postgres and MySQL because these two connectors only support to auto-generated the default id type ( integer in this case). 

From #3602 what suggests and real world usage, I am proposing to add `uuid` type to data type, and also add support for auto-generated string type id to Postgres connector:

### Proposal
- Similar to mongoDB, when property type is `uuid` and auto-generated, **by default,** use `uuid-ossp` extension and `uuid_generate_v4()` function to allow auto-generated string.
```
  @property({
    id: true,
    type: 'string',
    generated: true,
    useDefaultIdType: false, // this is needed if this property is id
    postgresql: {
      dataType: 'uuid',
    },
  })
  id: String;
```
~when id type is `string` and auto-generated, treat it as `uuid`. ( Should we?)~
- If user would like use other extension & function, they can do:
```
  @property({
    id: true,
    type: 'string',
    generated: true,
    useDefaultIdType: false, // this is needed if this property is id
    postgresql: {
      dataType: 'uuid',
      defaultFn: 'example_function',
      extension: 'my_extension'
    },
  })
  id: String;
```
However, it is user's responsibility to provide valid extension and function.
- If user would like to use other type as id type **and** make it auto-generated, should warn them that LoopBack doesn't support to auto-generate such type for Postgres. They need to do `ALTER TABLE` to set up the table manually.  
- Document the above setup rules, options, and limitations in both LB3 and LB4

If the above is okay, will fix MySQL as well.

## Checklist

### DIFF
--- a/lib/migration.js
+++ b/lib/migration.js
@@ -7,6 +7,7 @@
 const SG = require('strong-globalize');
 const g = SG();
 const async = require('async');
+const chalk = require('chalk');
 const debug = require('debug')('loopback:connector:postgresql:migration');
 
 module.exports = mixinMigration;
@@ -295,12 +296,33 @@ function mixinMigration(PostgreSQL) {
     const self = this;
     const modelDef = this.getModelDefinition(model);
     const prop = modelDef.properties[propName];
-    if (prop.generated) {
-      return 'SERIAL';
-    }
     let result = self.columnDataType(model, propName);
-    if (!self.isNullable(prop)) result = result + ' NOT NULL';
 
+    // checks if dataType is set to uuid
+    let postgDefaultFn;
+    let postgType;
+    const postgSettings = prop.postgresql;
+    if (postgSettings && postgSettings.dataType) {
+      postgType = postgSettings.dataType.toUpperCase();
+    }
+
+    if (prop.generated) {
+      if (result === 'INTEGER') {
+        return 'SERIAL';
+      } else if (postgType === 'UUID') {
+        if (postgSettings && postgSettings.defaultFn && postgSettings.extension) {
+          // if user provides their own extension and function
+          postgDefaultFn = postgSettings.defaultFn;
+          return result + ' NOT NULL' + ' DEFAULT ' + postgDefaultFn;
+        }
+        return result + ' NOT NULL' + ' DEFAULT uuid_generate_v4()';
+      } else {
+        console.log(chalk.red('>>> WARNING: ') +
+          `auto-generation is not supported for type "${chalk.yellow(prop.type)}". \
+          Please add your own function to the table "${chalk.yellow(model)}".`);
+      }
+    }
+    if (!self.isNullable(prop)) result = result + ' NOT NULL';
     result += self.columnDbDefault(model, propName);
     return result;
   };
@@ -313,32 +335,53 @@ function mixinMigration(PostgreSQL) {
   PostgreSQL.prototype.createTable = function(model, cb) {
     const self = this;
     const name = self.tableEscaped(model);
+    const modelDef = this.getModelDefinition(model);
+
+    // collects all extensions needed to be created
+    let createExtensions;
+    Object.keys(this.getModelDefinition(model).properties).forEach(function(propName) {
+      const prop = modelDef.properties[propName];
+
+      // checks if dataType is set to uuid
+      const postgSettings = prop.postgresql;
+      if (postgSettings && postgSettings.dataType && postgSettings.dataType === 'UUID'
+         && postgSettings.defaultFn && postgSettings.extension) {
+        createExtensions += 'CREATE EXTENSION IF NOT EXISTS "' + postgSettings.extension + '";';
+      }
+    });
+    // default extension
+    if (!createExtensions) {
+      createExtensions = 'CREATE EXTENSION IF NOT EXISTS "uuid-ossp";';
+    }
 
     // Please note IF NOT EXISTS is introduced in postgresql v9.3
-    self.execute('CREATE SCHEMA ' +
+    self.execute(
+      createExtensions +
+      'CREATE SCHEMA ' +
       self.escapeName(self.schema(model)),
-    function(err) {
-      if (err && err.code !== '42P06') {
-        return cb && cb(err);
-      }
-      self.execute('CREATE TABLE ' + name + ' (\n  ' +
-          self.propertiesSQL(model) + '\n)',
-      function(err, info) {
-        if (err) {
-          return cb(err, info);
+      function(err) {
+        if (err && err.code !== '42P06') {
+          return cb && cb(err);
         }
-        self.addIndexes(model, undefined, function(err) {
+        self.execute('CREATE TABLE ' + name + ' (\n  ' +
+          self.propertiesSQL(model) + '\n)',
+        function(err, info) {
           if (err) {
-            return cb(err);
+            return cb(err, info);
           }
-          const fkSQL = self.getForeignKeySQL(model,
-            self.getModelDefinition(model).settings.foreignKeys);
-          self.addForeignKeys(model, fkSQL, function(err, result) {
-            cb(err);
+          self.addIndexes(model, undefined, function(err) {
+            if (err) {
+              return cb(err);
+            }
+            const fkSQL = self.getForeignKeySQL(model,
+              self.getModelDefinition(model).settings.foreignKeys);
+            self.addForeignKeys(model, fkSQL, function(err, result) {
+              cb(err);
+            });
           });
         });
-      });
-    });
+      },
+    );
   };
 
   PostgreSQL.prototype.buildIndex = function(model, property) {
@@ -481,7 +524,7 @@ function mixinMigration(PostgreSQL) {
       default:
       case 'String':
       case 'JSON':
-        return 'TEXT';
+      case 'Uuid':
       case 'Text':
         return 'TEXT';
       case 'Number':
@@ -645,6 +688,7 @@ function mixinMigration(PostgreSQL) {
       case 'CHARACTER':
       case 'CHAR':
       case 'TEXT':
+      case 'UUID':
         return 'String';
 
       case 'BYTEA':
--- a/test/postgresql.migration.test.js
+++ b/test/postgresql.migration.test.js
@@ -14,7 +14,7 @@ describe('migrations', function() {
   before(setup);
 
   it('should run migration', function(done) {
-    db.automigrate('UserDataWithIndexes', done);
+    db.automigrate(['UserDataWithIndexes', 'OrderData', 'DefaultUuid'], done);
   });
 
   it('UserDataWithIndexes should have correct indexes', function(done) {
@@ -73,6 +73,42 @@ describe('migrations', function() {
       done();
     });
   });
+
+  it('OrderData should have correct prop type uuid with custom generation function', function(done) {
+    checkColumns('OrderData', function(err, cols) {
+      assert.deepEqual(cols, {
+        ordercode:
+        {column_name: 'ordercode',
+          column_default: 'uuid_generate_v1()',
+          data_type: 'uuid'},
+        ordername:
+        {column_name: 'ordername',
+          column_default: null,
+          data_type: 'text'},
+        id:
+        {column_name: 'id',
+          column_default: 'nextval(\'orderdata_id_seq\'::regclass)',
+          data_type: 'integer'},
+      });
+      done();
+    });
+  });
+
+  it('DefaultUuid should have correct id type uuid and default function v4', function(done) {
+    checkColumns('DefaultUuid', function(err, cols) {
+      assert.deepEqual(cols, {
+        defaultcode:
+        {column_name: 'defaultcode',
+          column_default: 'uuid_generate_v4()',
+          data_type: 'uuid'},
+        id:
+        {column_name: 'id',
+          column_default: 'nextval(\'defaultuuid_id_seq\'::regclass)',
+          data_type: 'integer'},
+      });
+      done();
+    });
+  });
 });
 
 function setup(done) {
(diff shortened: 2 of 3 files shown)


lib/migration.js
```diff
@@ -3,60 +3,69 @@
 // This file is licensed under the Artistic License 2.0.
 // License text available at https://opensource.org/licenses/Artistic-2.0
 
-'use strict';
-const SG = require('strong-globalize');
+"use strict";
+const SG = require("strong-globalize");
```
bajtos: Please fix your editor/IDE settings to use single quotes `'` and revert all changes in quoting style.

bajtos:
This is tricky issue, thank you @agnes512 for digging deep into this problem space.

IIUC, there are two aspects we need to address:

1. How to represent the database-generated UUID values inside LoopBack: how to define such model property, what JavaScript type it uses at runtime, how is it described in OpenAPI spec.

2. How to auto-migrate models using database-generated UUID value as the primary key.

Also IIUC, this issue is specific to SQL databases that typically use a numeric primary key when the value is generated by the database. For example, database-generated UUID primary keys are the default option in MongoDB.

While I like the idea of introducing a new data type `uuid` (see the bottom of my comment), I am reluctant to use that as the solution here, because such change affects many consumers of LoopBack definition language. Besides updating SQL connectors, we also need to update NoSQL connectors (including those we don't maintain ourselves), code converting LoopBack models to Swagger/OpenAPI, etc.

I find this situation similar to ObjectID issues we are facing in MongoDB - we have a string-like value that must be treated differently by the connector.

I prefer to find a solution where the property type stays as `string` and the UUID aspect is described via connector-specific metadata (if needed). For example:

```ts
@model()
export class Customer extends Entity {
  @property({
    type: 'string',
    id: true,
    useDefaultIdType: false,
    postgresql: {
      defaultFn: 'gen_random_uuid()'
    }
  })
  custid: string;
}
```

Or perhaps we can enhance `generated ` field to support non-boolean values:

- `generated: false` (or `generated ` not present): auto-generation is disabled
- `generated: true`: the default database mechanism is used
- `generated: 'uuid'`: values are generated as UUID values using database-specific features

To be honest, I feel both options described above are adding too much complexity and not enough benefits for the majority of our users. Let's try to find something simpler.

I like the following items in your original proposal:

> - When id type is string and auto-generated, treat it as uuid.
> - If user would like to use other type as id type and make it auto-generated, should throw an error and suggest them to remove generated: true, then do ALTER TABLE to set up the table manually.
> - When id is auto-generated and the type is not db default type, alert that user should modify the table to enable the auto generation themselves.

I would like to propose a slightly different solution building on your ideas:

- When id type is `string` and auto-generated, treat it as uuid (using database-provided mechanism).
- When id is auto-generated and we don't know how to auto-generate the value (e.g. because the database does not support auto-generated values for this particular data type), then alert that user should modify the table to enable the auto generation themselves.

In other words, if the user defines an primary key as auto-generated, then LoopBack should use the most sensible option supported by the database (i.e. `gen_random_uuid()` for PostgreSQL) or alert the user when there is no such option available/implemented by the connector.

IMO, the migration should not abort the process by throwing an error,  I am proposing to print a warning instead and continue with the migration process. I think that even if don't configure all auto-generation rules in the database, it is still very useful to have the rest of the schema created by the LoopBack migration tool. (We also don't want to abort migration in the middle, when some of the tables were already migrated/updated but some haven't.)

We can also apply these new rules to connectors that don't support auto-generated numeric values (MongoDB and Cloudant), see e.g. https://github.com/strongloop/loopback-next/issues/2040. Because schema migration is usually not executed for NoSQL databases, this would require the check to be performed at runtime. To keep things simple, it may be best to leave this part out of scope of this pull request, just keep it in mind.

---

Circling back to the proposal to add `uuid` type. I think it's a good idea to eventually implement `uuid` type as a first-class data type supported by all layers of LoopBack, from database persistence to JSON Schema validation in the REST API layer. This will make it easier to use `ObjectID` type in MongoDB, define JSON Schema validations to enforce uuid formatting of the string values, etc. There is a also a TC39 proposal to add `uuid` support to JavaScript standard library, see https://github.com/tc39/proposal-uuid, see especially the [FAQ](https://github.com/tc39/proposal-uuid#faq) section explaining why most application should always use v4 UUID algorithm only.

I am proposing to open a new feature request in loopback-next to implement `uuid` data type. Maybe a new Epic and a spike story to kick-off the work with the necessary research?

agnes512:
@bajtos thank you for elaborating all your concern!

> because such change affects many consumers of LoopBack definition language. Besides updating SQL connectors, we also need to update NoSQL connectors (including those we don't maintain ourselves), code converting LoopBack models to Swagger/OpenAPI, etc.

You're right, I didn't consider this part. Will just use the existing types.

> Or perhaps we can enhance generated field to support non-boolean values:
```
- `generated: false` (or `generated ` not present): auto-generation is disabled
- `generated: true`: the default database mechanism is used
- `generated: 'uuid'`: values are generated as UUID values using database-specific features
```

I was thinking to have something similar to mongoDB: 
```
    postgresql: {
      dataType: 'uuid'
    }
```
so when `generated` &&  `postgresql.dataType` === 'uuid` -> use `uuid_generate_v4()` as **default** function.

Any preferences?

bajtos:
> was thinking to have something similar to mongoDB:

```
    postgresql: {
      dataType: 'uuid'
    }
```

That's a decent solution too!

> Any preferences?

Personally, I would try to keep this simple and easy to use for our users. If their model/property definition says that the PK/`id` property should be an auto-generated string, then I would like the framework to fulfill that wish in the most sensible way. If we don't know how to do that or there isn't any sensible default, then notify the user about the problem.

agnes512:
@slnode test please

agnes512:
@slnode test please

agnes512:
@slnode test please

agnes512:
@slnode test please

agnes512:
@slnode test please

agnes512:
@slnode test please

emonddr:
Good discussion.

@agnes512 , this seems like a good feature. Great work!

agnes512:
@slnode test please

lib/migration.js
```diff
@@ -295,12 +296,33 @@ function mixinMigration(PostgreSQL) {
...
+      } else if (postgType === 'UUID') {
+        if (postgSettings && postgSettings.defaultFn && postgSettings.extension) {
+          // if user provides their own extension and function
+          postgDefaultFn = postgSettings.defaultFn;
+          return result + ' NOT NULL' + ' DEFAULT ' + postgDefaultFn;
+        }
+        return result + ' NOT NULL' + ' DEFAULT uuid_generate_v4()';
+      } else {
+        console.log(chalk.red('>>> WARNING: ') +
+            `auto-generattion is not supported for type "${chalk.yellow(prop.type)}". \
```
nabdelgadir: typo: `generattion` -> `generation`

lib/migration.js
```diff
@@ -295,12 +296,33 @@ function mixinMigration(PostgreSQL) {
...
+            Please add your own function to the table "${chalk.yellow(model)}".`);
```
nabdelgadir: What do you mean by `add your own function`?

lib/migration.js
```diff
@@ -313,32 +335,53 @@ function mixinMigration(PostgreSQL) {
   PostgreSQL.prototype.createTable = function(model, cb) {
     const self = this;
     const name = self.tableEscaped(model);
+    const modelDef = this.getModelDefinition(model);
+
+    // collects all extenstions need to be created
```
nabdelgadir: typo: `extenstions` -> `extensions`

lib/migration.js
```diff
@@ -313,32 +335,53 @@ function mixinMigration(PostgreSQL) {
...
-        return cb && cb(err);
-      }
-      self.execute('CREATE TABLE ' + name + ' (\n  ' +
-          self.propertiesSQL(model) + '\n)',
-      function(err, info) {
-        if (err) {
-          return cb(err, info);
+      function(err) {
+        if (err && err.code !== '42P06') {
+          return cb && cb(err);
```
nabdelgadir: Why return `cb` and not just `cb(err)`?

test/postgresql.migration.test.js
```diff
@@ -73,6 +73,42 @@ describe('migrations', function() {
       done();
     });
   });
+
+  it('OrderData should have correct prop type uuid with custom generation function', function(done) {
```
nabdelgadir: test title should be able to be read like a sentence and also be generic (not specific to `OrderData`) e.g. `it('checks that the correct prop type uuid with custom generation function is used`) or something

agnes512: I use the same format of previous test. Do you want me to change that one as well?

lib/migration.js
agnes512: functions of generating default values?

lib/migration.js
agnes512: I didn't change this part.  ( review it with whitespace hided

nabdelgadir: Oh sorry I refreshed, sounds good 👍

test/postgresql.migration.test.js
nabdelgadir: It's okay then. I don't think it's important for this PR.

lib/migration.js
```diff
@@ -295,12 +296,33 @@ function mixinMigration(PostgreSQL) {
...
+    if (postgSettings && postgSettings.dataType) {
+      postgType = postgSettings.dataType.toUpperCase();
+    }
+
     if (prop.generated) {
-      return 'SERIAL';
+      if (result === 'INTEGER') {
+        return 'SERIAL';
+      } else if (postgType === 'UUID') {
+        if (postgSettings && postgSettings.defaultFn && postgSettings.extension) {
```
jannyHou: Hmm, how do we deal with the `extension`? Is it a function or just a boolean value?

lib/migration.js
```diff
@@ -295,12 +296,33 @@ function mixinMigration(PostgreSQL) {
     const self = this;
     const modelDef = this.getModelDefinition(model);
     const prop = modelDef.properties[propName];
+    let result = self.columnDataType(model, propName);
```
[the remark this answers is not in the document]
jannyHou: sorry never mind, I just realized the code is in a function called `buildColumnDefinition`, so `result` is better.

lib/migration.js
```diff
@@ -481,7 +524,7 @@ function mixinMigration(PostgreSQL) {
       default:
       case 'String':
       case 'JSON':
-        return 'TEXT';
+      case 'Uuid':
```
jannyHou: Hmm, what's the difference between `Uuid` and `UUID`?

test/postgresql.migration.test.js
```diff
@@ -73,6 +73,42 @@ describe('migrations', function() {
...
+      done();
+    });
+  });
+
+  it('DefaultUuid should have correct id type uuid and default function v4', function(done) {
+    checkColumns('DefaultUuid', function(err, cols) {
+      assert.deepEqual(cols, {
+        defaultcode:
+        {column_name: 'defaultcode',
+          column_default: 'uuid_generate_v4()',
```
jannyHou: IIUC the `column_default` here is `uuid_generate_v4()` instead of `uuid_generate_v1()` because `extension` is missing in the [property def](https://github.com/strongloop/loopback-connector-postgresql/pull/404/files?utf8=%E2%9C%93&diff=split&w=1#diff-a55dd5264ec27c23b0c70c696af975eeR171), so I am thinking, maybe add a warning before [line 318](https://github.com/strongloop/loopback-connector-postgresql/pull/404/files?utf8=%E2%9C%93&diff=split&w=1#diff-546064e86fc7d116e6952283c43b8069R318) to let people know it falls back to the default value due to missing extension?

agnes512: Correct, it should  use the default extension and function for this case. And I have `uuid_generate_v4()`b here :smile:

lib/migration.js
agnes512: We need to 'install' the extension to use its functions. e.g, to use foo.bar(), we need to have foo installed first. In here extension is just a name that will be used in postgres query of creating table later.

lib/migration.js
agnes512: I am following the pattern with other type such as `String`, `Number`. But I think the case doesn't matter. I can change it to `uuid`.
As for the `UUID` below, it's modified to upper case so I use `UUID` there.

hacksparrow:
We need tests to assert the behavior when `useDefaultIdType` is not set and when the value is set to `true`.

agnes512:
> We need tests to assert the behavior when `useDefaultIdType` is not set and when the value is set to `true`.

I think `useDefaultIdType` is being tested in https://github.com/strongloop/loopback-datasource-juggler/pull/1783 ? Cause when `uuid` is set, the lb property will be set to `string`.

lib/migration.js
jannyHou: oh I see, the code comparison gave me an impression that they are two cases in the same function, while actually not lol. 👍

shreyanshgoel:
Great work people. But still I think we have a problem.
```
"my_id": {
	"id": true,
	"type": "string",
	"required": false,
	"generated": true,
	"useDefaultIdType": false,
	"postgresql": {
		"dataType": "uuid",
		"extenstion": "uuid-ossp",
		"defaultFn": "uuid_generate_v4()"
	}
},
```
So everything works well but when I try to do find on any model, I get id as NaN. Guessing because loopback is still treating my id as a number. This happens only when generated is true so i guess `useDefaultIdType` is no where being used. I check the whole code, could not find this parameter being user any where. I am using loopback 3.27.0.

Thank you, any update would be much apprecitated. 
Cheers

PR #405, opened 2019-12-11
uuid

agnes512:
fixes https://github.com/strongloop/loopback-next/issues/2398
Duplicate of https://github.com/strongloop/loopback-connector-postgresql/pull/404 because Jenkins fails randomly...

**_please review it with whitespace hided_**

strongloop/loopback-next#2398 is partially fixed in https://github.com/strongloop/loopback-datasource-juggler/pull/1783. 
```
@model()
export class Customer extends Entity {
  @property({
    type: 'string',
    id: true,
    useDefaultIdType: false,
    generated: true,
  })
  custid: string;
``` 

### Issue
The flag `useDefaultIdType: false` is supposed to allow to use custom type of auto-generated id property. But it only allows the default type, and it won't generated the custom id type itself as different DBs have different syntax/setup.

For example, Postgres and MySQL default identity type is INT. And they can be auto-generated by `SERIAL` and `AUTO_INCREAMENT` respectively. 

The model definition above won't work on Postgres and MySQL because these two connectors only support to auto-generated the default id type ( integer in this case). 

From #3602 what suggests and real world usage, I am proposing to add `uuid` type to data type, and also add support for auto-generated string type id to Postgres connector:

### Proposal

- Similar to mongoDB, when property type is `uuid` and auto-generated, **by default,** use `uuid-ossp` extension and `uuid_generate_v4()` function to allow auto-generated string.
```
  @property({
    id: true,
    type: 'string',
    generated: true,
    useDefaultIdType: false, // this is needed if this property is id
    postgresql: {
      dataType: 'uuid',
    },
  })
  id: String;
```
~when id type is `string` and auto-generated, treat it as `uuid`. ( Should we?)~
- If user would like use other extension & function, they can do:
```
  @property({
    id: true,
    type: 'string',
    generated: true,
    useDefaultIdType: false, // this is needed if this property is id
    postgresql: {
      dataType: 'uuid',
      defaultFn: 'example_function',
      extension: 'my_extension'
    },
  })
  id: String;
```
However, it is user's responsibility to provide valid extension and function.
- If user would like to use other type as id type **and** make it auto-generated, should warn them that LoopBack doesn't support to auto-generate such type for Postgres. They need to do `ALTER TABLE` to set up the table manually.  
- Document the above setup rules, options, and limitations in both LB3 and LB4

If the above is okay, will fix MySQL as well.

## Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -438,6 +438,57 @@ If there is a reference to an object being deleted then the `DELETE` will fail. 
 
 **Note**: The order of table creation is important. A referenced table must exist before creating a foreign key constraint.
 
+For **LoopBack 4** users, define your models under `models/` folder as follows:
+
+`customer.model.ts`:
+
+```ts
+@model()
+export class Customer extends Entity {
+  @property({
+    id: true,
+    type: 'Number',
+    required: false,
+    length: 20
+  })
+  id: number;
+
+  @property({
+    type: 'string',
+    length: 20
+  })
+  name: string;
+}
+```
+`order.model.ts`:
+
+```ts
+@model()
+export class Order extends Entity {
+  @property({
+    id: true,
+    type: 'Number',
+    required: false,
+    length: 20
+  })
+  id: number;
+
+  @property({
+    type: 'string',
+    length: 20
+  })
+  name: string;
+
+  @property({
+    type: 'Number',
+    length: 20
+  })
+  customerId: number;
+}
+```
+
+For **LoopBack 3** users, you can define your model JSON schema as follows:
+
 ```json
 {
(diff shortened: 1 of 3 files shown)


agnes512:
@slnode test please

agnes512:
@slnode test please

agnes512:
@slnode test please

agnes512:
@slnode test please

agnes512:
@slnode test please

rmg:
@slnode test please

agnes512:
@slnode test please

PR #416, opened 2020-02-13
fix Postgres random CI failure

agnes512:
Connects to https://github.com/strongloop/loopback-datasource-juggler/issues/1816

After investigating, the random CI failures are caused when running autoupdate, the race condition might cause the problem.
**__OR__** 
It failed to drop table (`customer_test2` for example) while doing autouodate, so that it failed to _update_ `customer_test2` because the relation still exists. 

I've simplified the test case to avoid the second situation:

### Before:
  Model `Order` has a foreign key `customerId` refers Model `C2`
-(autoupdate)-> 
  Model `Order` has a foreign key `customerId` refers Model `C3`
-(autoupdate)-> 
Model `Order` has fk `customerId` refers Model `C2` and fk `productId` refers to Model `Product`
-(autoupdate)-> 
Model `Order` doesn't have any fks.

### Simplified
  Model `Order` has created a foreign key `customerId` refers Model `C2`
-(autoupdate)-> 
Model `Order` has being updated to have a fk `productId` refers to Model `Product`
-(autoupdate)-> 
Model `Order` removed all fks.

I think the simplified one also checks that it should create, update, and delete foreign keys.

## Checklist

### DIFF
--- a/test/postgresql.autoupdate.test.js
+++ b/test/postgresql.autoupdate.test.js
@@ -296,8 +296,6 @@ describe('autoupdate', function() {
                     keys: ['firstname'],
                     order: ['ASC']},
                 });
-
-                // console.log(err, result);
                 done(err, result);
               });
             });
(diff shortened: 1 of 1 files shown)


test/postgresql.autoupdate.test.js
```diff
@@ -419,77 +417,7 @@ describe('autoupdate', function() {
...
-          },
-          'description': {
-            'type': 'String',
-            'required': false,
-            'length': 40,
-          },
-        },
-      };
-
-      const orderTest_schema_v2 = {
```
jannyHou: very very nitpick(😓 ): I find `orderTest_schema_v1,3,4` but not `orderTest_schema_v2`, can we rename 3 to 2, and 4 to 3 to have a consistent number?

test/postgresql.autoupdate.test.js
```diff
@@ -643,6 +552,7 @@ describe('autoupdate', function() {
         }
 
         // do initial update/creation of table with fk
+        ds.createModel(product_schema.name, product_schema.properties, product_schema.options);
```
jannyHou: Thank you for providing the test flow, it's very helpful for reviewer. Could you also include it in the code comment? 

Like IIUC here should be:
Model Order has a foreign key customerId refers Model C2
Initial update/creation of table with fk

agnes512: Added

agnes512:
@slnode test please

agnes512:
@slnode test please

PR #417, opened 2020-02-17
Feature: setting disable default sort

Erikdegroot89:
This PR adds add a settings called `disableDefaultSort` for disabling the default order by `id` on all find queries without an order parameter.

This change's particular use case applies to any model that doesn't use `serial` id types. 

In my case, models with a `uuid` id type (filled with uuid v4 and no particular ordering) showed performance improvements because the uuid does not have to be included in the index.

This is mostly based on the same code for the Mongo DB connector: https://github.com/strongloop/loopback-connector-mongodb/pull/437


slnode:

test/postgresql.order.test.js
```diff
@@ -0,0 +1,65 @@
...
+const should = require('should');
+
+let db, PostWithDisableDefaultSort;
+
+describe('Order settings', function() {
+  before(function() {
+    db = global.getSchema();
+  });
+
+  before(function(done) {
```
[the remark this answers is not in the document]
Erikdegroot89: Thanks, I've corrected the test.

Erikdegroot89:
Hi Agnes512, thank you for your reply.
If I understand correctly you're suggesting that the default `order` by id should only occur when the ID is of the `serial` datatype (or maybe `number` 'regular' type) ? I think that sound like a practical middle ground.
I'd suggest we do both, adding an option to only auto-order numeric types. 
The option would then look something like this:
```
"db": {
...
"defaultSort": true (Default) / "numericIdOnly" / false,
...
}
```
Would that be acceptable?

agnes512:
@slnode test please

README.md
```diff
@@ -145,6 +145,10 @@ Check out [node-pg-pool](https://github.com/brianc/node-pg-pool) and [node postg
       <td>Boolean</td>
       <td>Whether to try SSL/TLS to connect to server</td>
     </tr>
+    <tr>
+      <td>defaultSort</td>
```
agnes512: I am not good at naming :grimacing: would like to hear from the team @strongloop/sq-lb-apex

agnes512:
@slnode test please

README.md
nabdelgadir: Maybe `defaultIdSort` since it only applies to the `id` field? But I'm not good at naming either 😅

README.md
```diff
@@ -145,6 +145,10 @@ Check out [node-pg-pool](https://github.com/brianc/node-pg-pool) and [node postg
...
+      <td>boolean / string</td>
```
nabdelgadir: Nitpick: `boolean / string` -> `Boolean/String`

README.md
```diff
@@ -145,6 +145,10 @@ Check out [node-pg-pool](https://github.com/brianc/node-pg-pool) and [node postg
...
+      <td>Set to <code>false</code> to disable default sorting on <code>id</code> column(s), set to <code>numericIdOnly</code> to only apply to ID's with a number type <code>id</code>.</td>
```
nabdelgadir: Nitpick: `column(s), set` -> `column(s). Set`
`apply to ID's` -> `apply to IDs`

test/postgresql.order.test.js
```diff
@@ -0,0 +1,224 @@
+// Copyright IBM Corp. 2013,2019. All Rights Reserved.
```
nabdelgadir: Nitpick: `2013,2019` -> `2020`

test/postgresql.order.test.js
```diff
@@ -0,0 +1,224 @@
...
+      },
+      {
+        defaultSort: 'numericIdOnly',
+      },
+    );
+
+    PostWithNumericStringDefaultSort = db.define(
+      'PostWithNumericStringDefaultSort',
+      {
+        id: {type: String, id: true},
```
nabdelgadir: Just wondering, would it be possible to have a database with mixed id types (e.g. both `String` and `Number`) and what would happen in that case when it's sorted?

Erikdegroot89: Hi @nabdelgadir , 
Good question! It is possible, though not very common AFAIK. 
By default it would be sorted, but once set to `numericIdOnly` it doesn't sort them. This is because the implementation checks if all ID types are `Number` ID types. If they aren't all `Number` types (for instance, one is a `String`), it won't sort.

Another way to go would be to only sort the ID's that _are_ a `Number` type, and skip any other type, but I think that might be messy.

README.md
Erikdegroot89: Thanks, corrected it.

README.md
Erikdegroot89: Thanks, corrected it.

test/postgresql.order.test.js
Erikdegroot89: Thanks, corrected it.

Erikdegroot89:
Thanks for your feedback @agnes512  & @nabdelgadir! I refactored `defaultSort` to `defaultIdSort` and applied the style corrections you suggested.

agnes512:
@slnode test please

test/postgresql.order.test.js
nabdelgadir: Thank you for the explanation!

test/postgresql.order.test.js
```diff
@@ -0,0 +1,228 @@
...
+
+    sql.should.be.equal('SELECT "id","title","content" FROM "public"."postwithdefaultIdSort" ORDER BY "id"');
+    done();
+  });
+
+  it('builds non-ordering base query', function(done) {
+    const res = db.connector.buildSelect('PostWithDisabledDefaultIdSort', {});
+    const sql = res.sql;
+
+    sql.should.be.equal('SELECT "id","title","content" FROM "public"."postwithdisableddefaultIdSort"');
```
nabdelgadir: The tests are failing now because `postwithdisableddefaultIdSort` should be `postwithdisableddefaultidsort`(similar thing for the below tests).

Erikdegroot89: Thanks, I adjusted the casing as suggested.

agnes512:
@slnode test please

agnes512:
@Erikdegroot89 The PR is ready to land. I will merge it after the commits are squashed, thanks!

Erikdegroot89:
Good to hear @agnes512 , I squashed the changes into two commits.

agnes512:
@slnode test please

agnes512:
@slnode test please

agnes512:
👏 great work! thank you ❤️

PR #418, opened 2020-02-18
default values added for new required columns when running autoupdate()

karanssj4:
Fixes #123

## Checklist

### DIFF
--- a/lib/migration.js
+++ b/lib/migration.js
@@ -161,7 +161,8 @@ function mixinMigration(PostgreSQL) {
         model, self.column(model, propName), actualFields,
       );
       if (!found && self.propertyHasNotBeenDeleted(model, propName)) {
-        sql.push('ADD COLUMN ' + self.addPropertyToActual(model, propName));
+        // add new columns with default values
+        sql.push('ADD COLUMN ' + self.addPropertyToActual(model, propName) + self.columnDbDefault(model, propName));
       }
     });
     if (sql.length > 0) {
--- a/test/postgresql.migration.test.js
+++ b/test/postgresql.migration.test.js
@@ -14,7 +14,7 @@ describe('migrations', function() {
   before(setup);
 
   it('should run migration', function(done) {
-    db.automigrate(['UserDataWithIndexes', 'OrderData', 'DefaultUuid'], done);
+    db.automigrate(['UserDataWithIndexes', 'OrderData', 'DefaultUuid', 'DefaultValueAfterColumnAdd'], done);
   });
 
   it('UserDataWithIndexes should have correct indexes', function(done) {
@@ -109,6 +109,32 @@ describe('migrations', function() {
       done();
     });
   });
+
+  it('should add default value for new required columns', async function() {
+    const DefaultValueAfterColumnAdd = await db.getModel('DefaultValueAfterColumnAdd');
+    await DefaultValueAfterColumnAdd.create({
+      name: 'name1',
+    });
+    await db.defineProperty('DefaultValueAfterColumnAdd', 'createdByAdmin', {
+      type: Boolean, required: true, default: true,
+    });
+    await db.defineProperty('DefaultValueAfterColumnAdd', 'birthDate', {
+      type: Date, required: true, default: '2020-02-18T16:50:24.746Z',
+    });
+    await db.defineProperty('DefaultValueAfterColumnAdd', 'pendingPeriod', {
+      type: Number, required: true, default: 10,
+    });
+    await db.autoupdate(['DefaultValueAfterColumnAdd']);
+    const res = await DefaultValueAfterColumnAdd.findOne();
+
+    assert.deepEqual(res.toJSON(), {
+      id: 1,
+      name: 'name1',
+      createdByAdmin: true,
+      birthDate: new Date('2020-02-18T16:50:24.746Z'),
+      pendingPeriod: 10,
+    });
+  });
 });
 
 function setup(done) {
@@ -172,6 +198,10 @@ function setup(done) {
       }},
   });
 
+  const DefaultValueAfterColumnAdd = db.define('DefaultValueAfterColumnAdd', {
+    name: String,
+  });
+
   done();
 }
 
@@ -230,3 +260,4 @@ function checkColumns(table, cb) {
     cb(err, cols);
   });
 }
+


slnode:

karanssj4:
@agnes512 npm test fails on jugglerv3/v4, if i just test this particular module, it passes everything

agnes512:
@karanssj4 yes v3/v4 sometimes fail randomly :( but don't worry about it in this PR. And thanks for the PR ! I will review it shortly.

agnes512:
@slnode test please

agnes512:
@karanssj4 I checked those failures. Actually they are causing by the changes. The tests in this repository are run, so are the tests in Juggler. 
If we take a closer look at the one of the falling test cases:
https://github.com/strongloop/loopback-datasource-juggler/blob/master/test/manipulation.test.js#L998
```
      Player.defineProperty('active', {
        type: Boolean,
        default: false,
      });
      await db.autoupdate('Player');

      // And updateOrCreate an existing record
      const found = await Player.updateOrCreate({id: created.id, name: 'updated'});
      should(found.toObject().active).be.oneOf([
        undefined, // databases supporting `undefined` value
        null, // databases representing `undefined` as `null`
      ]);
```
it's expecting `null` or `undefined`. That's why some v3/v4 tests fail. 

I am not sure if it was an intentional decision. But I think `active` in this case should be `false`. And this fix makes sense to me. @strongloop/loopback-maintainers could you please confirm?

jannyHou:
@karanssj4 Thank you for the patch! I think the juggler test expects the following behavior: 

When a new column with default value gets added to a database, the existing data will have a **null value**(null or undefined, depends on the db) instead of the default value for that new column.
And new data created omitting that column will have a default value.

So is there a way to follow ^ behaviour when create new columns in postgresql?

lib/migration.js
```diff
@@ -161,7 +161,8 @@ function mixinMigration(PostgreSQL) {
         model, self.column(model, propName), actualFields,
       );
       if (!found && self.propertyHasNotBeenDeleted(model, propName)) {
-        sql.push('ADD COLUMN ' + self.addPropertyToActual(model, propName));
+        // add new columns with default values
+        sql.push('ADD COLUMN ' + self.addPropertyToActual(model, propName) + self.columnDbDefault(model, propName));
```
jannyHou: If we don't specify default value for a property, should `self.columnDbDefault(model, propName)` be skipped?
Like:

```js
const columnDefaultValue = self.columnDbDefault(model,propName);
if (columnDefaultValue) {
    sql.push('ADD COLUMN ' + self.addPropertyToActual(model, propName) + self.columnDbDefault(model, propName));
} else {
   sql.push('ADD COLUMN ' + self.addPropertyToActual(model, propName))
}
```

agnes512: Yes, if there is no default value set, `self.columnDbDefault(model, propName)` returns `''`.

karanssj4: Yeah, it returns an empty string and doesn't impact the final SQL query

karanssj4:
@jannyHou default values are only added when you add a new required column in this PR. That's the desired behaviour.. right?

karanssj4:
I checked out https://github.com/strongloop/loopback-datasource-juggler/pull/1704 and 
https://github.com/strongloop/loopback-datasource-juggler/issues/1692

The test case was written so as to NOT return a default value if the database doesn't have it. Here, the database in question does have the default value (that's how Postgres handles it)

@bajtos @jannyHou  any thoughts on how should this be handled?

bajtos:
I think I was involved in past discussions about handling default values, I'll try to find them to refresh my memory and then review the proposal and discussion here. I am not sure if I'll find enough time this week, I'll aim to get it done by Tuesday 2020-02-25. (Feel free to ping me if I don't.)

test/postgresql.migration.test.js
```diff
@@ -230,3 +260,4 @@ function checkColumns(table, cb) {
     cb(err, cols);
   });
 }
+
```
bajtos: Extra EOL at EOF, please remove.

bajtos:
I reviewed the proposed code changes, they look reasonable to me :+1:

Regarding the failing test from juggler: this is tricky! As have been correctly wrote above, the test was written with the assumption that when a new optional column with a default value is added to an existing table, existing records will have `NULL` value in that column. Quoting from https://github.com/strongloop/loopback-datasource-juggler/issues/1692#issuecomment-478499046:

> * Let's say we have an existing table with some columns and applications that are already using this table.
> * Now we add a new column with the following properties:
>   
>   * Existing records are updated to set the new column to NULL (or undefined).
>   * In LB application, we want to set the new colum to a default value when no value was provided by the client. This is achieved by using `default` metadata in property definition.
> * When we run the updated application and query the database, the new default value is incorrectly applied on old records.

I think the scenario described above is still valid, we just need to find a way how to improve the test to work with connectors that are setting the newly added column to the default value for existing rows as part of `autoupdate`.

For example, we can rework the test to define the new property in two steps:
1. In the first step, the property is defined with no default value and we run `autoupdate` to update database schema.
2. In the second step, we add `default` value to the property definition BUT DO NOT update database schema.

Alternatively, we can look for better ways how to test the scenario we were fixing as part of the pull request which introduced the problematic test. Quoting from https://github.com/strongloop/loopback-datasource-juggler/issues/1692#issuecomment-478499046 again:

> In their project, they are using PersistedModel to access a datasource that's not a database. As a result, the model is describing values that are not always returned back by the connector.

As yet another alternative, we can modify the connector to configure SQL columns with the provided default value only when the property is also `required`, that may be the easiest way how to move this pull request forward.

I remember that LoopBack treats `required` and `default` fields differently than SQL, therefore it may be reasonable to configure `default` in database only for required properties.  Unfortunately, I don't remember details about the differences and the explanation of why LoopBack implements `default` properties in the current way.

karanssj4:
@bajtos here's my $0.02 on this. If someone wants a new column with null values, they wouldn't register a default value (at least in SQL).
To support that use case, we do have hooks and remote methods.

If the user doesn't wanna do that..then worst case scenario, they would make a new column without default values, run autoupdate, then add a default value, run autoupdate again

agnes512:
Similar issue https://github.com/strongloop/loopback-connector-mongodb/issues/564 (?)

karanssj4:
@bajtos @jannyHou any pointers on how to proceed finally? should I keep it as is and update juggler tests? should I keep it only for required default values?

> If someone wants a new column with null values, they wouldn't register a default value (at least in SQL).
> To support that use case, we do have hooks and remote methods.

do you agree with this?

test/postgresql.migration.test.js
```diff
@@ -109,6 +109,32 @@ describe('migrations', function() {
       done();
     });
   });
+
+  it('should add default value for new required columns', async function() {
+    const DefaultValueAfterColumnAdd = await db.getModel('DefaultValueAfterColumnAdd');
+    await DefaultValueAfterColumnAdd.create({
+      name: 'name1',
+    });
+    await db.defineProperty('DefaultValueAfterColumnAdd', 'createdByAdmin', {
```
jannyHou: `defineProperty` is a sync function I think, see [its definition](https://github.com/strongloop/loopback-datasource-juggler/blob/master/lib/datasource.js#L1017)

agnes512:
I can update Juggler's tests if needed :D

bajtos:
> But I don't know if ^ would violate the fix in PR strongloop/loopback-datasource-juggler#1704 :p Will respect @bajtos 's opinion on this.

I am afraid I won't be able to review these changes in the next 2 weeks. @hacksparrow IIRC, you were involved in recent changes related to handling of default values, can you please take a look at this pull request?

karanssj4:
can we open this again? @bajtos

ketansp:
Need to open this again. This is still an issue in 2024.

PR #422, opened 2020-03-05
tests: checks if column is discovered and mapped correctly

agnes512:
add a test case for Postgres for changes made in https://github.com/strongloop/loopback-connector/pull/171

- checks if the id column name is `demoId`
- checks if the instance can be generated properly with not-null id value 
## Checklist

### DIFF
--- a/test/postgresql.discover.test.js
+++ b/test/postgresql.discover.test.js
@@ -8,6 +8,7 @@ process.env.NODE_ENV = 'test';
 require('should');
 
 const assert = require('assert');
+const should = require('should');
 const _ = require('lodash');
 
 const DataSource = require('loopback-datasource-juggler').DataSource;
@@ -239,6 +240,36 @@ describe('Discover model properties', function() {
 });
 
 describe('Discover model primary keys', function() {
+  let Demo;
+  before(function(done) {
+    const demo_schema = {
+      'name': 'Demo',
+      'options': {
+        'idInjection': false,
+        'postgresql': {
+          'schema': 'public',
+          'table': 'demo',
+        },
+      },
+      'properties': {
+        'id': {
+          'type': 'Number',
+          'required': true,
+          'id': 1,
+          'postgresql': {
+            'columnName': 'demoId',
+          },
+        },
+        'name': {
+          'type': 'String',
+          'required': true,
+        },
+      },
+    };
+    Demo = db.createModel(demo_schema.name, demo_schema.properties, demo_schema.options);
+    Demo.destroyAll(done);
+  });
+
   it('should return an array of primary keys for product', function(done) {
     db.discoverPrimaryKeys('product', function(err, models) {
       if (err) {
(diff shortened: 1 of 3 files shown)


test/postgresql.discover.test.js
```diff
@@ -239,6 +240,36 @@ describe('Discover model properties', function() {
...
+          },
+        },
+        'name': {
+          'type': 'String',
+          'required': true,
+        },
+      },
+    };
+    Demo = db.createModel(demo_schema.name, demo_schema.properties, demo_schema.options);
+    Demo.destroyAll(done);
```
emonddr: Maybe add a comment. why create a model and then destroy all?

test/postgresql.discover.test.js
```diff
@@ -257,6 +288,33 @@ describe('Discover model primary keys', function() {
     });
   });
 
+  it('primary key should be discovered, and db generates instances properly', function(done) {
+    db.discoverPrimaryKeys('demo', function(err, models) {
+      if (err) {
+        console.error(err);
+        done(err);
+      } else {
+        models.forEach(function(m) {
```
emonddr: Why loop if there is only 1 model?

test/postgresql.discover.test.js
agnes512: `db.createModel` is to create model so that we can do CRUD operations. `.destroyAll` is to delete all instances. They are and basic steps and used in many test files.

test/postgresql.discover.test.js
agnes512: Match with other tests.

nabdelgadir: I think it makes sense to not loop if it's only one model, even if it's inconsistent with the other tests, but it's not a big deal.

test/postgresql.discover.test.js
```diff
@@ -257,6 +288,33 @@ describe('Discover model primary keys', function() {
...
+          assert.deepEqual(m, {owner: 'public',
+            tableName: 'demo',
+            columnName: 'demoId',
+            keySeq: 1,
+            pkName: 'demo_pkey'});
+        });
+      }
+    });
+    Demo.create({id: 1, name: 'checkCase'}, function(err) {
+      should.not.exists(err);
```
nabdelgadir: Can you change this and the one below to `should.not.exist(err)`? (https://shouldjs.github.io/#should-not-exist)

test/postgresql.discover.test.js
```diff
@@ -257,6 +288,33 @@ describe('Discover model primary keys', function() {
     });
   });
 
+  it('primary key should be discovered, and db generates instances properly', function(done) {
```
nabdelgadir: `should discover primary key and db generates instances properly`?

test/postgresql.discover.test.js
```diff
@@ -257,6 +288,33 @@ describe('Discover model primary keys', function() {
...
+            keySeq: 1,
+            pkName: 'demo_pkey'});
+        });
+      }
+    });
+    Demo.create({id: 1, name: 'checkCase'}, function(err) {
+      should.not.exists(err);
+      Demo.findOne({}, function(err, d) {
+        should.not.exists(err);
+        should.exists(d);
```
nabdelgadir: Similar thing here: `should.exists(d);` ->  `should.exist(d);` (https://shouldjs.github.io/#should-exist)

test/postgresql.discover.test.js
agnes512: edited, thanks!

agnes512:
@slnode test please

PR #427, opened 2020-03-20
Fix serialization of arrays of string in update queries values.

onaluf:
This pull request fixes a bug I encountered with the PostgeSQL connector when updating the value of a model's property of type array of strings.

When creating the model no issue occurs but when updating it the value stored in the column turns into an invalid array, for example `'{ "A","B","C"}'` instead of `'["A","B","C"]`. After a bit of investigating I  thought it was linked to the juggler giving the value in the query as an `Array` instead of a  `List` object but even after fixing this the issue was still there. So I figured that it must be in the connector and changed it there. 

If you think this change belong somewhere else just tell me and I'll move it.

The added unit test just check that the object is still deserialisable after an update. It fails without the change and pass with it.

## Checklist

👉 [Read and sign the CLA (Contributor License Agreement)](https://cla.strongloop.com/agreements/strongloop/loopback-connector-postgresql) 👈  (done) 

### DIFF
--- a/lib/postgresql.js
+++ b/lib/postgresql.js
@@ -766,6 +766,10 @@ PostgreSQL.prototype.toColumnValue = function(prop, val, isWhereClause) {
     });
   }
 
+  if (prop.type instanceof Array && prop.type[0] === String) {
+    return JSON.stringify(val);
+  }
+
   return val;
 };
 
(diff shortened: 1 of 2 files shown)


slnode:

Issue #342, opened 2018-09-05
Malformed Array Literal


sujeshthekkepatt:
Hai,
I am using sequelize for the migration task instead of the default automigrate function. I have created a column "tags" with datatype Array in sequelize.  And created the model using "lb model" and defined the type of tags as the array of string. But when i POST to /api/users it throws malformed array literal error. Turns out the default automigrate function create the type of "tags" to "text" not to actual postgres array. Any way to use the default Postgres array type in here?
My model definition is as,

{
  "name": "user",
  "plural": "users",
  "base": "mfy-base",
  "idInjection": false,
  "options": {
    "validateUpsert": true
  },
  "forceId": false,
  "properties": {
    "id": {
      "type": "string",
      "id": true,
      "required": false,
      "defaultFn": "uuid",
      "postgresql": {
        "dataType": "uuid"
      }
    },
    "psid": {
      "type": "string"
    },
    "profile_pic": {
      "type": "string"
    },
    "gender": {
      "type": "string"
    },
    "locale": {
      "type": "string"
    },
    "timezone": {
      "type": "string"
    },
    "status": {
      "type": "string",
      "default": "unread"
    },
    "tags": {
      "type": ["string"]
    },
    "last_message": {
      "type": "object"
    },
    "last_fb_message_at": {
      "type": "date",
      "defaultFn": "now"
    },
    "mmail": {
      "type": "string"
    },
    "is_mmail_allowed": {
      "type": "boolean",
      "default": true
    },
    "is_messaging_allowed": {
      "type": "boolean",
      "default": true
    },
    "first_name": {
      "type": "string"
    },
    "last_name": {
      "type": "string"
    },
    "type": {
      "type": "string"
    }
  },
  "validations": [],
  "relations": {
    "page": {
      "type": "belongsTo",
      "model": "page",
      "foreignKey": "",
      "options": {
        "nestRemoting": true
      }
    },
    "events": {
      "type": "hasMany",
      "model": "event",
      "foreignKey": "",
      "options": {
        "nestRemoting": true
      }
    }
  },
  "acls": [],
  "methods": {}
}

padelagani:
@sujeshthekkepatt 

Have you got any solution for this? I am also facing the same issue.

@bajtos Any update on this?

sujeshthekkepatt:
No. You may have to convert the type to 'text' at the database level. Since they are not doing any DB level operations it is very difficult to modify.

bajtos:
[...]

PR #428, opened 2020-03-20
Fix serialization of arrays of string in update

onaluf:
This pull request fixes a bug I encountered with the PostgeSQL connector when updating the value of a model's property of type array of strings.

When creating the model no issue occurs but when updating it the value stored in the column turns into an invalid array, for example '{ "A","B","C"}' instead of '["A","B","C"]. After a bit of investigating I thought it was linked to the juggler giving the value in the query as an Array instead of a List object but even after fixing this the issue was still there. So I figured that it must be in the connector and changed it there. (This may be related to #342)

If you think this change belong somewhere else just tell me and I'll move it.

The added unit test just check that the object is still deserialisable after an update. It fails without the change and pass with it.

## Checklist

👉 [Read and sign the CLA (Contributor License Agreement)](https://cla.strongloop.com/agreements/strongloop/loopback-connector-postgresql) 👈 (done)

### DIFF
--- a/lib/postgresql.js
+++ b/lib/postgresql.js
@@ -766,6 +766,24 @@ PostgreSQL.prototype.toColumnValue = function(prop, val, isWhereClause) {
     });
   }
 
+  if (Array.isArray(prop.type)) {
+    // There is two possible cases for the type of "val" as well as two cases for dataType
+    const isArrayDataType = prop.postgresql && prop.postgresql.dataType === 'varchar[]';
+    if (Array.isArray(val)) {
+      if (isArrayDataType) {
+        return val;
+      } else {
+        return JSON.stringify(val);
+      }
+    } else {
+      if (isArrayDataType) {
+        return JSON.parse(val);
+      } else {
+        return val;
+      }
+    }
+  }
+
   return val;
 };
 
--- a/test/postgresql.test.js
+++ b/test/postgresql.test.js
@@ -64,6 +64,15 @@ describe('postgresql connector', function() {
       loc: 'GeoPoint',
       created: Date,
       approved: Boolean,
+      tags: {
+        type: ['string'],
+      },
+      categories: {
+        type: ['string'],
+        postgresql: {
+          dataType: 'varchar[]',
+        },
+      },
     });
     created = new Date();
   });
@@ -199,6 +208,59 @@ describe('postgresql connector', function() {
         done();
       });
     });
+  });
+
+  it('should support creating and updating arrays with default dataType', function(done) {
+    let postId;
+    Post.create({title: 'Updating Arrays', content: 'Content', tags: ['AA', 'AB']})
+      .then((post)=> {
+        postId = post.id;
+        post.should.have.property('tags');
+        post.tags.should.be.Array();
+        post.tags.length.should.eql(2);
+        post.tags.should.eql(['AA', 'AB']);
+        return Post.updateAll({where: {id: postId}}, {tags: ['AA', 'AC']});
+      })
+      .then(()=> {
+        return Post.findOne({where: {id: postId}});
+      })
+      .then((post)=> {
+        post.should.have.property('tags');
+        post.tags.should.be.Array();
+        post.tags.length.should.eql(2);
+        post.tags.should.eql(['AA', 'AC']);
+        done();
+      })
+      .catch((error) => {
+        done(error);
+      });
+  });
+
+  it('should support creating and updating arrays with "varchar[]" dataType', function(done) {
+    let postId;
+    Post.create({title: 'Updating Arrays', content: 'Content', categories: ['AA', 'AB']})
+      .then((post)=> {
+        postId = post.id;
+        post.should.have.property('categories');
+        post.should.have.property('categories');
+        post.categories.should.be.Array();
+        post.categories.length.should.eql(2);
+        post.categories.should.eql(['AA', 'AB']);
+        return Post.updateAll({where: {id: postId}}, {categories: ['AA', 'AC']});
+      })
+      .then(()=> {
+        return Post.findOne({where: {id: postId}});
+      })
+      .then((post)=> {
+        post.should.have.property('categories');
+        post.categories.should.be.Array();
+        post.categories.length.should.eql(2);
+        post.categories.should.eql(['AA', 'AC']);
+        done();
+      })
+      .catch((error) => {
+        done(error);
+      });
   });
 
   it('should support boolean types with false value', function(done) {


slnode:

achrinza:
@slnode test please

onaluf:
Ok so I've seen the test that failed was that in the mean time my branch got behind master... I'll try to rebase it as soon as possible.

spirAde:
Hello, guys. I've also faced with the same problem and looked through the solution of the author of PR. Thank you @onaluf for your solution. I've checked your solution for my case and it doesn't work, sorry. I took next example(I consider only loopback4):

```
@property.array(String, {
    default: [SectionNames.PROFILE, SectionNames.EDUCATION, SectionNames.SKILL],
    postgresql: {
      columnName: 'sections_order',
      dataType: 'varchar[]',
      dataLength: null,
      dataPrecision: null,
      dataScale: null,
      nullable: 'NO',
    },
  })
  sectionsOrder: string[];
```

And got next error:

```
{
    "error": {
        "statusCode": 500,
        "name": "error",
        "message": "malformed array literal: \"\"[\\\"PROFILE\\\",\\\"EDUCATION\\\",\\\"SKILL\\\"]\"\"",
        "length": 182,
        "severity": "ERROR",
        "code": "22P02",
        "detail": "Array value must start with \"{\" or dimension information.",
        "file": "arrayfuncs.c",
        "line": "316",
        "routine": "array_in",
        "stack": "error: malformed array literal: \"\"[\\\"PROFILE\\\",\\\"EDUCATION\\\",\\\"SKILL\\\"]\"\"\n    at Connection.parseE (/home/spirade/WebstormProjects/resume/node_modules/pg/lib/connection.js:581:48)\n    at Connection.parseMessage (/home/spirade/WebstormProjects/resume/node_modules/pg/lib/connection.js:380:19)\n    at Socket.<anonymous> (/home/spirade/WebstormProjects/resume/node_modules/pg/lib/connection.js:116:22)\n    at Socket.emit (events.js:189:13)\n    at addChunk (_stream_readable.js:284:12)\n    at readableAddChunk (_stream_readable.js:265:11)\n    at Socket.Readable.push (_stream_readable.js:220:10)\n    at TCP.onStreamRead [as onread] (internal/stream_base_commons.js:94:17)"
    }
}
```

Also you don't consider number type, as I can see. I mean this kind of check:
`
if (prop.type instanceof Array && prop.type[0] === String) {
`

From my perspective, this solution is better: 
```
if (prop.type instanceof Array) {
  return `{${JSON.parse(val).join(', ')}}`;
}
```
and it works with number and strings, and after that data processes correctly.
Best wishes and have a nice day

test/postgresql.test.js
```diff
@@ -820,6 +820,53 @@ describe('Serial properties', function() {
...
+        },
+        function(callback) {
+          Model.create({productCodes: ['AA', 'AB']}, callback);
+        },
+        function(callback) {
+          Model.updateAll({where: {id: 1}}, {productCodes: ['AA', 'AC']}, callback);
+        },
+        function(callback) {
+          Model.findOne({where: {id: 1}}, function(err, r) {
+            r.should.have.property('productCodes');
```
jannyHou, line 861: please also add a check for `productCodes`'s value:

should.equal(r.productCodes, ['AA', 'AC']);

test/postgresql.test.js
```diff
@@ -820,6 +820,53 @@ describe('Serial properties', function() {
...
+
+  it('should return a valid array', function(done) {
+    const schema =
+      {
+        'name': 'TestWithArray',
+        'properties': {
+          'id': {
+            'type': 'number', 'id': true, 'generated': true,
+          },
+          'productCodes': {
```
jannyHou, line 838: Can you add `productCodes` as a property of the existing model `Post`?
See its definition in https://github.com/strongloop/loopback-connector-postgresql/blob/dc47d13d5e15d5c31d75fb7c044d2e97542072df/test/postgresql.test.js#L61-L66

And you can add the tests after this one(it also tests patch):
https://github.com/strongloop/loopback-connector-postgresql/blob/dc47d13d5e15d5c31d75fb7c044d2e97542072df/test/postgresql.test.js#L193

test/postgresql.test.js
```diff
@@ -820,6 +820,53 @@ describe('Serial properties', function() {
...
+    const count = 0;
+    Model.attachTo(db);
+
+    db.automigrate(function(err, data) {
+      async.series([
+        function(callback) {
+          Model.destroyAll(callback);
+        },
+        function(callback) {
+          Model.create({productCodes: ['AA', 'AB']}, callback);
```
jannyHou, line 854: We support the promise chain of CRUD operations, example see 
https://github.com/strongloop/loopback-datasource-juggler/blob/master/test/crud-with-options.test.js#L149

So you can write the chain calls as 

```js
Post.create(params)
  .then(()=> {
     return Post.updateAll(params)
   })
  .then(()=> {
      // other function calls
   })
```

spirAde:
Hello, guys. This solution isn't work, because it's not real postgresql array type in database. It will be convert just as text field type in database. You can easily check this, if you add postgresql settings in productCodes field
```
'productCodes': {
  'type': ['string'],
  'postgresql': {
    'dataType': 'varchar[]',
  }
},
```

After that the code of author will fail with error from pg: malformed array literal: ""[\"AA\",\"AB\"]"", because postgresql array type expects {'AA', 'AB'}.

onaluf:
@jannyHou I've changed the test according to your pointers, hope it suits you.

@spirAde yes the array is saved as a string in the DB but it's then deserialized correctly. I think this is the correct behavior, at least that's the current way the connector behaves when creating the entry in the first place.... only the updating is problematic as far as I can tell.

onaluf:
@spirAde Sorry I think I now get what you mean... if the dataType is explicitly set to `varchar[]` then it fails... however by default the array is saved as a string in the db... I'll try to see how to accommodate both cases.

onaluf:
Ok @spirAde, @jannyHou I think this time I've got it. I've written the test for both cases and generalized the solution for non-string array. 

The issue, as far as I understand it, was that `toColumnValue` received different type of values for arrays depending on the situation... on top of that the expected value varied based on the `dataType` of the field:
- Creating an object always ended with val being a `string` (of a standard stringified array)
- Updating an object always ended with val being an `Array`
- Default `dataType` wanted a `string`
- Array `dataType` wanted an `Array`

This is why I could create but not update my model AND people in #342 could update but not create theirs.

I hope the proposed solution is suitable but don't hesitate and I'll continue to improve it if needed.

lib/postgresql.js
```diff
@@ -766,6 +766,26 @@ PostgreSQL.prototype.toColumnValue = function(prop, val, isWhereClause) {
     });
   }
 
+  if (prop.type instanceof Array) {
+    // There is two possible cases for the type of "val" as well as two cases for dataType
+    const isArrayDataType = prop.postgresql &&
+                            prop.postgresql.dataType &&
+                            (prop.postgresql.dataType.indexOf('[]') > -1);
```
onaluf, line 773: There may be a more elegant way to detect this but I couldn't find one...

spirAde:
@onaluf thank you, it works. Awesome

test/postgresql.test.js
```diff
@@ -201,6 +210,50 @@ describe('postgresql connector', function() {
     });
   });
 
+  it('should support creating and updating arrays with default dataType', function(done) {
+    let postId;
+    Post.create({title: 'Updating Arrays', content: 'Content', tags: ['AA', 'AB']})
+      .then((post)=> {
+        postId = post.id;
+        post.should.have.property('tags');
```
jannyHou, line 218: Nitpick: You can assert the value for `tags` directly as
```js
 post.should.have.property('tags');
 post.tags.should.be.Array();
 post.tags.length.should.eql(2);
 post.tags.should.eql(['AA', 'AB']);
```
so first check property exists, then its type, then its length(optional), then the value. 
Usually these guarantee if test fails, the error message will point to the exact wrong place :)

test/postgresql.test.js
```diff
@@ -201,6 +210,50 @@ describe('postgresql connector', function() {
...
+        post.tags[1].should.eql('AB');
+        return Post.updateAll({where: {id: postId}}, {tags: ['AA', 'AC']});
+      })
+      .then((wooot)=> {
+        return Post.findOne({where: {id: postId}});
+      })
+      .then((post)=> {
+        post.should.have.property('tags');
```
jannyHou, line 226: nitpick: same here

```js
post.should.have.property('tags');
 post.tags.should.be.Array();
 post.tags.length.should.eql(2);
 post.tags.should.eql(['AA', 'AC']);
```

test/postgresql.test.js
```diff
@@ -201,6 +210,50 @@ describe('postgresql connector', function() {
...
+  it('should support creating and updating arrays with "varchar[]" dataType', function(done) {
+    let postId;
+    Post.create({title: 'Updating Arrays', content: 'Content', categories: ['AA', 'AB']})
+      .then((post)=> {
+        postId = post.id;
+        post.should.have.property('categories');
+        post.categories[1].should.eql('AB');
+        return Post.updateAll({where: {id: postId}}, {categories: ['AA', 'AC']});
+      })
+      .then((wooot)=> {
```
jannyHou, line 244: ditto, leaving it empty is ok :)

test/postgresql.test.js
```diff
@@ -201,6 +210,50 @@ describe('postgresql connector', function() {
...
+        return Post.findOne({where: {id: postId}});
+      })
+      .then((post)=> {
+        post.should.have.property('categories');
```
jannyHou, line 248: ditto

```js
post.should.have.property('tags');
 post.tags.should.be.Array();
 post.tags.length.should.eql(2);
 post.tags.should.eql(['AA', 'AC']);
```

test/postgresql.test.js
```diff
@@ -201,6 +210,50 @@ describe('postgresql connector', function() {
...
+        done(error);
+      });
+  });
+
+  it('should support creating and updating arrays with "varchar[]" dataType', function(done) {
+    let postId;
+    Post.create({title: 'Updating Arrays', content: 'Content', categories: ['AA', 'AB']})
+      .then((post)=> {
+        postId = post.id;
+        post.should.have.property('categories');
```
jannyHou, line 240: ditto
```js
post.should.have.property('tags');
 post.tags.should.be.Array();
 post.tags.length.should.eql(2);
 post.tags.should.eql(['AA', 'AB']);
```

jannyHou:
> I'm not sure about serialization of array and convert it to database text column type. 

According to function [buildColumnType](https://github.com/strongloop/loopback-connector-postgresql/blob/e118c490a06b218cce61d1a1f46488479b9e832e/lib/migration.js#L524), this is the expected behavior. 

So the array type data is serialized to a string then deserialized to the array.

I think the reason is SQL databases have weaker support for object, and the DSL of LB is inspired by mongodb...So type like array and object are serialized.

lib/postgresql.js
jannyHou: @onaluf is `varchar[]` the only possible data type? or datatype in format `*[]` are all valid conditions?

If `varchar[]` is the only valid one, you can use 
```js
const isArrayDataType = prop.postgresql &&
prop.postgresql.dataType === 'varchar[]';
```
instead.

lib/postgresql.js
```diff
@@ -766,6 +766,26 @@ PostgreSQL.prototype.toColumnValue = function(prop, val, isWhereClause) {
     });
   }
 
+  if (prop.type instanceof Array) {
```
jannyHou, line 769: Nitpick: `Array.isArray(prop.type)` is better

lib/postgresql.js
```diff
@@ -766,6 +766,26 @@ PostgreSQL.prototype.toColumnValue = function(prop, val, isWhereClause) {
...
+    // There is two possible cases for the type of "val" as well as two cases for dataType
+    const isArrayDataType = prop.postgresql &&
+                            prop.postgresql.dataType &&
+                            (prop.postgresql.dataType.indexOf('[]') > -1);
+    if (val instanceof Array) {
```
jannyHou, line 774: ditto `Array.isArray(val)` is better.

jannyHou:
@slnode test please

onaluf:
@jannyHou Thanks for the constructive feedback! I think I've solved all the point you raised.

jannyHou:
@slnode test please

jannyHou:
@onaluf Almost there, could you squash the 2 commits into 1? Then we will help you land it!

onaluf:
Done!

jannyHou:
@slnode test please

jannyHou:
@slnode test please

jannyHou:
Merged 🎉 Thanks everyone again for the contribution and feedback!

spirAde:
@jannyHou @onaluf Thank you, guys!

Issue #447, opened 2020-06-16
Impossible to exclude tables from migration


gioppoluca:
## Steps to reproduce

Execute migration on models that are only read only and on top of tables not managed by the project (but needed in the application)
If I execute a migration also these tables are taken into account

## Current Behavior

Try to migrate all the models

## Expected Behavior

Be able to exclude a model from the migration

gioppoluca:
I saw that it is possible to add a list of models to the "app.migrateSchema" but it should be better and more generic to have an option in the model definition

dhmlau:
Yes, it's possible to specify the model(s) that you want to run the database migration (also see code snippet in https://github.com/strongloop/loopback-next/issues/2831#issue-440068034). 

Having an option in the model definition to indicate this preference makes sense to me. It is probably related to the PR you submitted https://github.com/strongloop/loopback-connector-postgresql/pull/449. I'm thinking this should be implemented at an more abstract level rather than at the postgresql connector level itself. Let's continue the discussion on the PR.

dhmlau:
PR #449

PR #449, opened 2020-06-16
added optional schema

### LINKED ISSUE #446: Migration fails on existing schema different than public with user with no schema creation ACL
## Steps to reproduce

- Have a PostgreSQL Db with an existing schema (thiis is needed since we need to add permissions also to view other tables in other schemas etc)
- Create a model with the schema set to the existing schema
- Execute the npm run migrate and get error on permission denied
## Current Behavior

Migrating schemas (alter existing schema)
Cannot migrate database schema { error: permission denied for database micado
    at Connection.parseE (/code/micado-backend/node_modules/pg/lib/connection.js:581:48)
    at Connection.parseMessage (/code/micado-backend/node_modules/pg/lib/connection.js:380:19)
    at Socket.<anonymous> (/code/micado-backend/node_modules/pg/lib/connection.js:116:22)
    at Socket.emit (events.js:198:13)
    at addChunk (_stream_readable.js:288:12)
    at readableAddChunk (_stream_readable.js:269:11)
    at Socket.Readable.push (_stream_readable.js:224:10)
    at TCP.onStreamRead [as onread] (internal/stream_base_commons.js:94:17)
  name: 'error',
  length: 97,
  severity: 'ERROR',
  code: '42501',
  detail: undefined,
  hint: undefined,
  position: undefined,
  internalPosition: undefined,
  internalQuery: undefined,
  where: undefined,
  schema: undefined,
  table: undefined,
  column: undefined,
  dataType: undefined,
  constraint: undefined,
  file: 'aclchk.c',
  line: '3485',
  routine: 'aclcheck_error' }

## Expected Behavior

The migration operates with no error

## Additional information

I've seen that it seems that the connector try to CREATE the schema, but the schema already exists and the operating user has no right to create a schema in the DB.

If this is the case we should need an option NOT to create the schema if this is existing OR add the "if not exists" option so that the operation goes on normally
## Related Issues

_See [Reporting Issues](http://loopback.io/doc/en/contrib/Reporting-issues.html) for more tips on writing good issues_

gioppoluca:
This pull add an optional settings (migrationCreateSchema) in the datasource allowing the user to decide if he wants to create the schema or not.
This will allow the user to use migration on already existing schemas in PostgreSQL

Fixes #446

## Checklist

- [x ] `npm test` passes on your machine
- [x ] Code conforms with the [style guide](https://loopback.io/doc/en/contrib/style-guide-es6.html)
- [x ] Commit messages are following our [guidelines](https://loopback.io/doc/en/contrib/git-commit-messages.html)

### DIFF
--- a/lib/migration.js
+++ b/lib/migration.js
@@ -354,11 +354,13 @@ function mixinMigration(PostgreSQL) {
       createExtensions = 'CREATE EXTENSION IF NOT EXISTS "uuid-ossp";';
     }
 
+    let createSchema = (typeof self.settings.migrationCreateSchema !== 'undefined') ?  self.settings.migrationCreateSchema : true
+
     // Please note IF NOT EXISTS is introduced in postgresql v9.3
     self.execute(
       createExtensions +
-      'CREATE SCHEMA ' +
-      self.escapeName(self.schema(model)),
+      (createSchema ? 'CREATE SCHEMA ' +
+      self.escapeName(self.schema(model)): ''),
       function(err) {
         if (err && err.code !== '42P06') {
           return cb && cb(err);


slnode:

agnes512:
> This will allow the user to use migration on already existing schemas in PostgreSQL

IIUC, do you mean to drop schema objects if they already exist and re-create them based on model definitions?

gioppoluca:
> > This will allow the user to use migration on already existing schemas in PostgreSQL
> 
> IIUC, do you mean to drop schema objects if they already exist and re-create them based on model definitions?

No.
As I explained in the issue the bug is that, in the migration code, the connector ALWAYS try to issue a "CREATE SCHEMA" command.
There could be cases where the SCHEMA is already been created outside the Loopback application because Loopback only manages part of the DB or for any good reasons.
In those cases we need the connector NOT to try to execute that statement for the following reasons:
1) it require the user of the datasource to have CREATE role on the DB and that could not be the case. I have a DB with 5 schemas and the user that access the schema where the tables managed by Loopback DO NOT have CREATE role on the DB and DO NOT WANT to give that privilege to him just because the Loopback connector issues a useless "CREATE SCHEMA" command.
In this way we can decide if we want to have the connector create the schema or not (only during migration obviously).
2) is a useless operation (in fact it fails with error, but PostgreSQL let you go on with the operation) if the schema is already there
3) is a sub-optimal piece of code since also in the comment there is the specification that the "IF NOT EXISTS" is available only from a certain version of PostgreSQL so is a known useless operation in some cases
4) it does an operation that is not directly linked to the tables: creation of schema should be done apart and been an explicit decision of the user and not done as a consequence of a property of a table

gioppoluca:
To add some more info the example of the datasource config follows:

`{
  "name": "nameDS",
  "connector": "postgresql",
  "url": "postgres://xxx:yyyy@host/db",
  "host": "hostdb",
  "port": 5432,
  "user": "unpriviledged_user",
  "password": "pwd",
  "schema": "not_public_schema",
  "migrationCreateSchema": false,
  "database": "db"
}`

agnes512:
Thanks for the detailed explanations.

Here are some of my thoughts:
`CREATE SCHEMA` is not that useful, agreed, as the schema already exists in the most of the cases. However I think `CREATE SCHEMA` will still need to be part of the migration process. For that reason, even I am not a big fan of adding flags, I couldn't come up a better solution.

My concerned is, other operations might require different kind of permissions too( [creating extension](https://github.com/strongloop/loopback-connector-postgresql/blob/master/lib/migration.js#L359) for example), should we add a flag for that case? Also, what if the schema does not exist and the user doesn't have the permission?

I am okay with fix but at the same time would it make more sense if we just let database (or LB) warn about the permission?

dhmlau:
@gioppoluca, thanks for the PR.

Cross posting my comment in https://github.com/strongloop/loopback-connector-postgresql/issues/447#issuecomment-645522738:
> Having an option in the model definition to indicate this preference makes sense to me. It is probably related to the PR you submitted #449. I'm thinking this should be implemented at an more abstract level rather than at the postgresql connector level itself. Let's continue the discussion on the PR.

@agnes512 @jannyHou @hacksparrow @bajtos, thoughts?

agnes512:
>  I'm thinking this should be implemented at an more abstract level rather than at the postgresql connector level itself.

+1

Adding a flag might work for this specific case, but there still are some cases we need to consider. I am leaning towards letting the database decide which operations can be run by the user.

> I saw that it is possible to add a list of models to the "app.migrateSchema" but it should be better and more generic to have an option in the model definition

Personally I'd like to have an option for the migrate script that which models are being updated/creared, something like:`npm run migrate --model User`

Cause for me the migration preferences is out the scope of the model definition.

gioppoluca:
Sorry it should be better not to mix topics.
This PR is only for the management of the schema in the migration.
Have you decided on this specific proposal?
At the moment my solution will be not to use Loopback for the migration since it does not do all the needed operations and also require to have users with permissions that are higher then the required and this could be a security problem for my case.
Thanks

bajtos:
Instead of adding a new option, can we modify the SQL/DDL script to detect if the schema already exists and don't try to create it in such case? 

#446 mentions _add the "if not exists" option so that the operation goes on normally_, would that solve the issue? 

If not, then perhaps we can call [`this.discoverDatabaseSchemas()`](https://github.com/strongloop/loopback-connector/blob/7d43b461d13e2e67fd07cee4ae22c1e19d0ea848/lib/sql.js#L1739-L1747) to get the list of all schemas so that we can check if our target schema already exists.

dhmlau:
We just switch the contribution method from CLA to DCO, making your contribution easier in the future. Please sign the commits with DCO by amending your commit messages with `-s` flag and push the changes again. If you're using the command line, you can:

```
git commit --amend -s
git push --force-with-lease
```

Please refer to [this docs page](https://loopback.io/doc/en/contrib/code-contrib.html#developer-certificate-of-origin-dco) for details. Thanks! 

For more questions, please join our [Slack workspace - #loopback-contributors](https://loopbackio.slack.com/join/shared_invite/zt-8lbow73r-SKAKz61Vdao~_rGf91pcsw#/).

PR #458, opened 2020-09-02
Add on delete options on FK constraints

QuentinLB:
Allow setting `on delete` rule on foreign key constraints.
Update foreign key discovery query to add on delete rules mainly for testing purpose.

## Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -713,6 +713,8 @@ export class Customer extends Entity {
         entity: 'Customer',
         entityKey: 'id',
         foreignKey: 'customerId',
+        onDelete: 'CASCADE',
+        onUpdate: 'SET NULL'
       },
     },
   })
@@ -764,7 +766,9 @@ export class Order extends Entity {
         "name": "fk_order_customerId",
         "entity": "Customer",
         "entityKey": "id",
-        "foreignKey": "customerId"
+        "foreignKey": "customerId",
+        "onDelete": "CASCADE",
+        "onUpdate": "SET NULL"
       }
     }
   },
@@ -787,7 +791,7 @@ export class Order extends Entity {
 </details>
 <br>
 {% include tip.html content="
-Removing or updating the value of `foreignKeys` will be updated or delete or update the constraints in the db tables. If there is a reference to an object being deleted then the `DELETE` will fail. Likewise if there is a create with an invalid FK id then the `POST` will fail.
+Removing or updating the value of `foreignKeys` will be updated or delete or update the constraints in the db tables. If there is a reference to an object being deleted then the `DELETE` will fail. Likewise if there is a create with an invalid FK id then the `POST` will fail. The `onDelete` and `onUpdate` properties are optional and will default to `NO ACTION`.
 " %}
 
 ### Auto-generated ids
--- a/lib/migration.js
+++ b/lib/migration.js
@@ -574,7 +574,9 @@ function mixinMigration(PostgreSQL) {
           const fkRefTable = self.table(fkEntityName);
           needsToDrop = !isCaseInsensitiveEqual(fkCol, fk.fkColumnName) ||
                         !isCaseInsensitiveEqual(fkRefKey, fk.pkColumnName) ||
-                        !isCaseInsensitiveEqual(fkRefTable, fk.pkTableName);
+                        !isCaseInsensitiveEqual(fkRefTable, fk.pkTableName) ||
+                        parseAction(newFk.onDelete) != fk.onDelete ||
+                        parseAction(newFk.onUpdate) != fk.onUpdate;
         } else {
           // FK will be dropped if column is removed
           // only if FK is in model properties then need to drop
@@ -636,12 +638,28 @@ function mixinMigration(PostgreSQL) {
       // verify that the other model in the same DB
       if (this._models[fkEntityName]) {
         return 'CONSTRAINT ' + this.escapeName(fk.name) + ' ' +
-          'FOREIGN KEY (' + this.escapeName(fk.foreignKey) + ') ' +
-          'REFERENCES ' + this.tableEscaped(fkEntityName) + '(' + fk.entityKey + ')';
+        'FOREIGN KEY (' + this.escapeName(fk.foreignKey) + ') ' +
+        'REFERENCES ' + this.tableEscaped(fkEntityName) + '(' + fk.entityKey + ') ' +
+        'ON DELETE ' + parseAction(fk.onDelete) + ' ' +
+        'ON UPDATE ' + parseAction(fk.onUpdate);
       }
     }
     return '';
   };
+
+  /*!
+   * Process model settings foreign key action,
+   * if action is not a valid sql action return 'NO ACTION'
+   * @param {Any} action
+   */
+  function parseAction(action) {
+    if (typeof action !== 'string') return 'NO ACTION';
+    const _action = action.toUpperCase();
+    if (['RESTRICT', 'CASCADE', 'SET NULL', 'SET DEFAULT'].includes(_action))
+      return _action;
+    else
+      return 'NO ACTION';
+  }
 
   /*!
    * Case insensitive comparison of two strings
--- a/test/postgresql.autoupdate.test.js
+++ b/test/postgresql.autoupdate.test.js
@@ -469,6 +469,8 @@ describe('autoupdate', function() {
               'entity': 'Product',
               'entityKey': 'id',
               'foreignKey': 'productId',
+              'onDelete': 'CASCADE',
+              'onUpdate': 'SET NULL',
             },
           },
         },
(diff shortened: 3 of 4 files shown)


agnes512 (commented):
Thanks for the PR! Could you remove the `yarn.lock` file? Apart from that, could you also update the README file? If you don't know where is the best for it, I can do it after the PR gets merged 👍 

I am also wondering if we should support other rules such as `restrict` as well as `onUpdate`, see similar code in https://github.com/strongloop/loopback-connector-mysql/blob/master/lib/migration.js#L40 . We can open follow-up issues later.

agnes512:
@slnode test please

QuentinLB:
Hello @agnes512 

I updated the README and removed the yarn.lock file, the commit linter (didn't know it was a thing) seems to have a problem with my last commit but when I click on "Details" it asks me to authorize my github account somewhere I don't know. 

Regarding the other rules, I agree but my PR is more a proof of concept than anything. I needed cascading delete on my current project and I thought it would be more productive to share it with the maintainers, maybe open the discussion. I deliberately restricted my addition to `ON DELETE CASCADE` rule to avoid side effects on unimplemented rules.

I browsed the mysql connector you linked, it looks like not much more work to implement the whole thing but as of this week I won't be able to do more work on this PR. Next week I'll be able to spend few hours on it.

lib/migration.js
```diff
@@ -636,8 +636,9 @@ function mixinMigration(PostgreSQL) {
       // verify that the other model in the same DB
       if (this._models[fkEntityName]) {
         return 'CONSTRAINT ' + this.escapeName(fk.name) + ' ' +
-          'FOREIGN KEY (' + this.escapeName(fk.foreignKey) + ') ' +
-          'REFERENCES ' + this.tableEscaped(fkEntityName) + '(' + fk.entityKey + ')';
+        'FOREIGN KEY (' + this.escapeName(fk.foreignKey) + ') ' +
+        'REFERENCES ' + this.tableEscaped(fkEntityName) + '(' + fk.entityKey + ')' +
+        (fk.onDelete && fk.onDelete.toUpperCase() === 'CASCADE' ? ' ON DELETE CASCADE' : '');
```
bajtos, line 641: Have you considered adding support for other `ON DELETE` modes? Quoting from https://www.postgresql.org/docs/8.2/ddl-constraints.html:

> Restricting and cascading deletes are the two most common options. `RESTRICT` prevents deletion of a referenced row. `NO ACTION` means that if any referencing rows still exist when the constraint is checked, an error is raised; this is the default behavior if you do not specify anything. (The essential difference between these two choices is that `NO ACTION` allows the check to be deferred until later in the transaction, whereas `RESTRICT` does not.) `CASCADE` specifies that when a referenced row is deleted, row(s) referencing it should be automatically deleted as well. There are two other options: `SET NULL` and `SET DEFAULT`. These cause the referencing columns to be set to nulls or default values, respectively, when the referenced row is deleted. Note that these do not excuse you from observing any constraints. For example, if an action specifies `SET DEFAULT` but the default value would not satisfy the foreign key, the operation will fail.

README.md
```diff
@@ -721,9 +723,11 @@ export class Order extends Entity {
 
 </details>
 <br>
-{% include tip.html content="
-Removing or updating the value of `foreignKeys` will be updated or delete or update the constraints in the db tables. If there is a reference to an object being deleted then the `DELETE` will fail. Likewise if there is a create with an invalid FK id then the `POST` will fail.
-" %}
+
+Tips: 
+* Removing or updating the value of `foreignKeys` will be updated or delete or update the constraints in the db tables. If there is a reference to an object being deleted then the `DELETE` will fail. Likewise if there is a create with an invalid FK id then the `POST` will fail.
+* The `onDelete` property is optional, if set only `CASCADE` is properly supported right now.
```
bajtos, line 729:
```suggestion
* The `onDelete` property is optional, only the following values are supported now: `CASCADE`.
```

test/postgresql.autoupdate.test.js
```diff
@@ -578,6 +579,7 @@ describe('autoupdate', function() {
               assert.equal(foreignKeys[0].pkTableName, 'customer_test2');
               assert.equal(foreignKeys[0].fkColumnName, 'customerId');
               assert.equal(foreignKeys[0].fkName, 'fk_ordertest_customerId');
+              assert.equal(foreignKeys[0].onDelete, 'NO ACTION');
```
bajtos, line 582: If our auto-discovery is setting `onDelete` to `NO ACTION`, then I think `autoupdate`/`automigrate` must support `NO ACTION` too. Otherwise it's not possible to migrate/update models that were auto-discovered.

lib/migration.js
bajtos, line 641: We need to notify the user when they are using unsupported `onDelete` configuration. I see two options:

1. Log a warning via `console.warn`, continue the migration using default `ON DELETE` behavior.
2. Throw an error to abort auto-migration, force the user to pick a supported `ON DELETE` behavior.

Personally, I prefer tighter checks to avoid the situation where the user accidentally misses a warning and then complains to us about `onDelete` flag not working, therefore I would choose the second option (throw an error).

bajtos (changes requested):
Thank you @QuentinLB for the pull request, I love the idea of adding support for `ON DELETE` configuration of (foreign key) constraints.

I would like to discuss the implementation details, see below.

bajtos:
> Regarding the other rules, I agree but my PR is more a proof of concept than anything. I needed cascading delete on my current project and I thought it would be more productive to share it with the maintainers, maybe open the discussion. I deliberately restricted my addition to `ON DELETE CASCADE` rule to avoid side effects on unimplemented rules.

Makes sense :+1: Personally, I am fine to support only few `ON DELETE` options in the first version. (My comment https://github.com/strongloop/loopback-connector-postgresql/pull/458#discussion_r483470867 can be considered as resolved then.)

However, I'd like to avoid the situation where auto-discovery creates model that cannot be updated/migrated later, so I think we need to support at least `NO ACTION` - see https://github.com/strongloop/loopback-connector-postgresql/pull/458#discussion_r483471760.

> I browsed the mysql connector you linked, it looks like not much more work to implement the whole thing but as of this week I won't be able to do more work on this PR. Next week I'll be able to spend few hours on it.

Don't worry! It's fine to work in small increments. Perhaps we can start by adding `ON DELETE` support in this pull request, and then look into adding more features in follow-up pull requests later, as you have time?

bajtos:
@slnode test please

test/postgresql.autoupdate.test.js
QuentinLB: Make sense, but there's an issue how to distinguish when user explicitly set `onDelete` to `NO ACTION` or when he didn't set anything and PG did it as default ?

For this PR it's not an issue since only `CASCADING` is fully supported but if more rules are supported this will need to be handled.

Is removing this assertion a suitable answer for the scope of this PR ?

lib/migration.js
QuentinLB: I agree, personally I prefer to throw an error as it will force user to fix the issue, ensure that nothing is inconsistent and have a low probability to break existing code base since it's a new feature.

bajtos: I think we are in agreement then :)

test/postgresql.autoupdate.test.js
bajtos: > Make sense, but there's an issue how to distinguish when user explicitly set `onDelete` to `NO ACTION` or when he didn't set anything and PG did it as default ?

Do we need to distinguish between that? 

> For this PR it's not an issue since only `CASCADING` is fully supported but if more rules are supported this will need to be handled.
> 
> Is removing this assertion a suitable answer for the scope of this PR ?

I am not sure, how is autodiscovery working at the moment? Will it set `onDelete` to `'NO ACTION'` or to `undefined`?

I don't want to expand scope of this PR more than needed, but I feel that we must support migration of models that were defined using the auto-discovery mechanism.

bajtos: Interesting. If I am reading our auto-discovery implementation correctly, then we don't support discovering `onDelete` behavior for foreign keys.

https://github.com/strongloop/loopback-connector-postgresql/blob/ced6c286f7baffaf8f569af747d943d0da96ffa4/lib/discovery.js#L262-L284

If that's the case, then we can close this discussion as resolved.

However, it would be great if you could update README to mention that auto-discovery does not support `onDelete` rules yet.

QuentinLB:
Hi everyone, long time no see. 

Yesterday I was migrating databases and I noticed that my patch to support `ON DELETE CASCADE` constraints wasn't working anymore. I fix the issue and took the opportunity to implement the full spec also with `ON UPDATE` actions. 

Sorry about the commit formatting and missing DCO, I forgot it was a thing with your repo, I did though signed the previous auth system you used to mandate.

agnes512:
Thank you @QuentinLB ! I've rebased the PR and squashed commits. It's good to go 😄

QuentinLB:
Thanks a lot, I can finally come back to the main repo :smiley:

PR #460, opened 2020-09-04
array_type_field_filter: Add new contains filter to search on Array type field

### LINKED ISSUE #342: Malformed Array Literal

shubhisood:
- Adds new filter named as contains.
- This filter will search on Array type fields in Postgres DB.

Include references to all related GitHub issues and other pull requests, for example:

Fixes #342 

## Checklist

- [X] DCO (Developer Certificate of Origin) [signed in all commits](https://loopback.io/doc/en/contrib/code-contrib.html)
- [X] `npm test` passes on your machine
- [X] New tests added or existing tests modified to cover all changes
- [X] Code conforms with the [style guide](https://loopback.io/doc/en/contrib/style-guide-es6.html)
- [X] Commit messages are following our [guidelines](https://loopback.io/doc/en/contrib/git-commit-messages.html)

### DIFF
--- a/lib/postgresql.js
+++ b/lib/postgresql.js
@@ -522,6 +522,9 @@ PostgreSQL.prototype.buildExpression = function(columnName, operator,
       const regexOperator = operatorValue.ignoreCase ? ' ~* ?' : ' ~ ?';
       return new ParameterizedSQL(columnName + regexOperator,
         [operatorValue.source]);
+    case 'contains':
+      return new ParameterizedSQL(columnName + ' @> array[' + operatorValue.map((v) => `'${v}'`) + ']::'
+        + propertyDefinition.postgresql.dataType);
     default:
       // invoke the base implementation of `buildExpression`
       return this.invokeSuper('buildExpression', columnName, operator,
--- a/test/postgresql.test.js
+++ b/test/postgresql.test.js
@@ -73,7 +73,7 @@ describe('postgresql connector', function() {
           dataType: 'varchar[]',
         },
       },
-    });
+    }, {allowExtendedOperators: true});
     created = new Date();
   });
 
@@ -264,15 +264,23 @@ describe('postgresql connector', function() {
   });
 
   it('should support where filter for array type field', async () => {
-    Post.dataSource.settings.allowExtendedOperators = true;
-    const post = await Post.find({where: {and: [
+    await Post.create({
+      title: 'LoopBack Participates in Hacktoberfest',
+      categories: ['LoopBack', 'Announcements'],
+    });
+    await Post.create({
+      title: 'Growing LoopBack Community',
+      categories: ['LoopBack', 'Community'],
+    });
+
+    const found = await Post.find({where: {and: [
       {
-        categories: {'contains': ['AA']},
+        categories: {'contains': ['LoopBack', 'Community']},
       },
     ]}});
-    should.exist(post);
-    post.length.should.equal(1);
+    found.map(p => p.title).should.deepEqual(['Growing LoopBack Community']);
   });
+
   it('should support boolean types with false value', function(done) {
     Post.create(
       {title: 'T2', content: 'C2', approved: false, created: created},
--- a/test/postgresql.test.js
+++ b/test/postgresql.test.js
@@ -263,6 +263,16 @@ describe('postgresql connector', function() {
       });
   });
 
+  it('should support where filter for array type field', async () => {
+    Post.dataSource.settings.allowExtendedOperators = true;
+    const post = await Post.find({where: {and: [
+      {
+        categories: {'contains': ['AA']},
+      },
+    ]}});
+    should.exist(post);
+    post.length.should.equal(1);
+  });
   it('should support boolean types with false value', function(done) {
     Post.create(
       {title: 'T2', content: 'C2', approved: false, created: created},
--- a/test/postgresql.test.js
+++ b/test/postgresql.test.js
@@ -218,7 +218,7 @@ describe('postgresql connector', function() {
         post.should.have.property('tags');
         post.tags.should.be.Array();
         post.tags.length.should.eql(2);
-        post.tags.should.eql(['AA', 'AB']);
+        post.tags.toArray().should.eql(['AA', 'AB']);
         return Post.updateAll({where: {id: postId}}, {tags: ['AA', 'AC']});
       })
       .then(()=> {
@@ -228,7 +228,7 @@ describe('postgresql connector', function() {
         post.should.have.property('tags');
         post.tags.should.be.Array();
         post.tags.length.should.eql(2);
-        post.tags.should.eql(['AA', 'AC']);
+        post.tags.toArray().should.eql(['AA', 'AC']);
         done();
       })
       .catch((error) => {
@@ -245,7 +245,7 @@ describe('postgresql connector', function() {
         post.should.have.property('categories');
         post.categories.should.be.Array();
         post.categories.length.should.eql(2);
-        post.categories.should.eql(['AA', 'AB']);
+        post.categories.toArray().should.eql(['AA', 'AB']);
         return Post.updateAll({where: {id: postId}}, {categories: ['AA', 'AC']});
       })
       .then(()=> {
@@ -255,7 +255,7 @@ describe('postgresql connector', function() {
         post.should.have.property('categories');
         post.categories.should.be.Array();
         post.categories.length.should.eql(2);
-        post.categories.should.eql(['AA', 'AC']);
+        post.categories.toArray().should.eql(['AA', 'AC']);
         done();
       })
       .catch((error) => {
--- a/README.md
+++ b/README.md
@@ -45,7 +45,7 @@ const config = {
 
 <details><summary markdown="span"><strong>For LoopBack 3 users</strong></summary>
 
-Use the [Data source generator](http://loopback.io/doc/en/lb3/Data-source-generator.html) to add a PostgreSQL data source to your application.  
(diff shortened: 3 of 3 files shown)


test/postgresql.test.js
```diff
@@ -263,6 +263,18 @@ describe('postgresql connector', function() {
       });
   });
 
+  it('should support where filter for array type field', function(done) {
+    Post.dataSource.settings.allowExtendedOperators = true;
```
shubhisood, line 267: After adding this we do not need juggler related PR.
PFB PR Link:
https://github.com/strongloop/loopback-datasource-juggler/pull/1865

test/postgresql.test.js
```diff
@@ -263,6 +263,18 @@ describe('postgresql connector', function() {
...
+    Post.find({where: {and: [
+      {
+        categories: {'@>': ['AA']},
+      },
+    ]}}, function(err, post) {
```
bajtos, line 272: I find this filter too complicated, would the following simpler version work too?

```js
Post.find({where: {categories: {'@>': ['AA']}}})
```

test/postgresql.test.js
```diff
@@ -263,6 +263,18 @@ describe('postgresql connector', function() {
...
+      should.not.exist(err);
```
bajtos, line 273:
```suggestion
      if (err) return done(err);
```

That way the original error details and the stack trace is preserved.

test/postgresql.test.js
```diff
@@ -263,6 +263,18 @@ describe('postgresql connector', function() {
       });
   });
 
+  it('should support where filter for array type field', function(done) {
```
bajtos, line 266: We prefer to use async/await style for all newly written tests. Can you please rewrite your test as follows?

```js
it('should support where filter for array type field', async () => {
  const post = await Post.find(/* your filter */);
  should.exist(post);
});
```

lib/postgresql.js
```diff
@@ -522,6 +522,9 @@ PostgreSQL.prototype.buildExpression = function(columnName, operator,
       const regexOperator = operatorValue.ignoreCase ? ' ~* ?' : ' ~ ?';
       return new ParameterizedSQL(columnName + regexOperator,
         [operatorValue.source]);
+    case '@>':
```
bajtos, line 525: I find the operator name `@>` as very cryptic. Have you considered using a more descriptive name like `contains`? 

Having said that, maybe the operator `@>` is something that PostgreSQL users are familiar with, and a new name like `contains` would just add learning overhead?

Let's discuss.

bajtos (commented):
I like how it's possible to implement your desired feature inside the connector only :+1:

I have few comments on implementation details, PTAL above.

Please add a new section to README to explain what extended operators are supported (e.g. `@>`) and what is needed to allow the app to use them (set `allowExtendedOperators` in model and/or datasource settings).

lib/postgresql.js
bajtos: Another aspect to consider: this operator can be used in REST API clients too, e.g. to call 

```http
GET https://somehost/api/posts?filter={"where":{"category":{"@>":["AA"]}}}
```

From the point of view of REST API consumers (e.g. front-end developers), I think a descriptive operator name like `contains` is much better.

```http
GET https://somehost/api/posts?filter={"where":{"category":{"contains":["AA"]}}}
```

test/postgresql.test.js
shubhisood: Yes, @bajtos it is working fine too.

test/postgresql.test.js
shubhisood: Done with this test-case change

lib/postgresql.js
shubhisood: Done renamed operator from @> to contains

test/postgresql.test.js
shubhisood: How we would be handling this with async await type test case? @bajtos 
I was thinking to write test case this way for handling failure of testcase with proper details.


it('should support where filter for array type field', async () => {
      Post.dataSource.settings.allowExtendedOperators = true;
      try {
        const post = await Post.find({where: {
          categories: {'contains': ['AA']},
        }});
        should.exist(post);
        post.length.should.equal(1);
      } catch (err) {
        should.not.exist(err);
      }
    });


Please tell me if it is correct, I will update current test case in this way only.

bajtos: As a rule of thumb, the tests should pass any unexpected error to the test runner:
- Use `done(err)` when using callback
- Use regular error propagation when using async functions

```js
it('should support where filter for array type field', async () => {
      Post.dataSource.settings.allowExtendedOperators = true;
        const post = await Post.find({where: {
          categories: {'contains': ['AA']},
        }});
        should.exist(post);
        post.length.should.equal(1);
    });
```

bajtos (commented):
The new version looks pretty good.

1. Please change the commit message to use the new operator name

2. Please add a new section to README to explain what extended operators are supported (e.g. `@>`) and what is needed to allow the app to use them (set `allowExtendedOperators` in model and/or datasource settings). Put yourself in the shoes of other LoopBack users - where would you look for information about `contains` operator and what info would you want to see to help you understand how to use this feature in your app?

shubhisood:
@bajtos Thanks for review..
For Point#1 I have done the update.
For point number 2, I have following queries:
In which section of readme should I add this operator, As in current readme I couldn't find currently supported operators of postgres in this connector.

bajtos:
> In which section of readme should I add this operator, As in current readme I couldn't find currently supported operators of postgres in this connector.

That's a good question. I believe the postgres connector so far supports only built-in LoopBack operators supported by all (or most?) connectors, so there was no need to explicitly list extended operators.

Now that you are implementing the first extended operator, I think you should simply add a new section. You can draw inspiration from MongoDB connector README:

- The section [Additional properties](https://github.com/strongloop/loopback-connector-mongodb/blob/4551cc0428a6954091981eb8e4861e7c1c9b37d8/README.md#additional-properties) mentions the configuration setting `allowExtendedOperators` and the extended operators unlocked by this option
- The section [Update operators](https://github.com/strongloop/loopback-connector-mongodb/blob/4551cc0428a6954091981eb8e4861e7c1c9b37d8/README.md#update-operators) adds more details

In the case of PostgreSQL and your new `contains` operator, we can either call the section something like `Query operators` or perhaps a more generic `Extended operators`.

shubhisood:
> > In which section of readme should I add this operator, As in current readme I couldn't find currently supported operators of postgres in this connector.
> 
> That's a good question. I believe the postgres connector so far supports only built-in LoopBack operators supported by all (or most?) connectors, so there was no need to explicitly list extended operators.
> 
> Now that you are implementing the first extended operator, I think you should simply add a new section. You can draw inspiration from MongoDB connector README:
> 
> * The section [Additional properties](https://github.com/strongloop/loopback-connector-mongodb/blob/4551cc0428a6954091981eb8e4861e7c1c9b37d8/README.md#additional-properties) mentions the configuration setting `allowExtendedOperators` and the extended operators unlocked by this option
> * The section [Update operators](https://github.com/strongloop/loopback-connector-mongodb/blob/4551cc0428a6954091981eb8e4861e7c1c9b37d8/README.md#update-operators) adds more details
> 
> In the case of PostgreSQL and your new `contains` operator, we can either call the section something like `Query operators` or perhaps a more generic `Extended operators`.

Added the readme section for the new contains operator

bajtos:
Thank you @shubhisood for the update. I pushed few more commits to fix the failing builds and improve README content. Let's wait for CI results before landing.

bajtos (approved):

bajtos:
@slnode test please

bajtos:
@slnode test please

bajtos:
Landed, thank you @shubhisood for the contribution ❤️

shubhisood:
> Landed, thank you @shubhisood for the contribution

Thanks for merging @bajtos

PR #469, opened 2020-11-28
Fix bug multicolumn indexes dropped on autoupdate

### LINKED ISSUE #464: Multi-Column Database Indexes are Always Dropped on Autoupdate
## Steps to reproduce

1. Create a multi-column index on a database model's schema. E.g. for a model, `person`, in model's definition JSON file:
```javascript
...
"indexes" {
  "unique_name_index": {
    "keys": { "firstName": 1, "lastName": 1 },
    "options": { "unique": true }
  }
},
...
```
2. Execute `ds.autoupdate()` where `ds` is the database data source
3. Verify `unique_name_index` exists on the `person` table
4. Execute `ds.autoupdate()` a second time 
## Current Behavior
Despite the fact that the index definition hasn't changed, `loopback-connector-postgresql` will still `DROP` the index for `unique_name_index` and recreate it. Obviously this is not great for performance as the table's data needs to be re-indexed even though the index hasn't changed.

## Expected Behavior
`loopback-connector-postgresql` detects that `unique_name_index` exists and it's definition (i.e. `keys`, `type`, and `unique`-ness properties) have not changed, so the connector doesn't `DROP` the index.

Given that the index is recreated during the `autoupdate()` call, it's not easy to detect that the index is first dropped and then re-created. We discovered the issue when we externally added a partial index (since `loopback-connector-postgresql` doesn't support partial indexes) and attempted to have Loopback "skip" dropping and re-creating the index by defining an index on the model's schema in Loopback with the same name, keys, type, and uniqueness. We found that even though our index matched all the properties `loopback-connector-postgresql` uses to determine if an existing index in the database matches the model schema's index definition, the index was still dropped and then recreated using the model schema's definition (i.e. it dropped our partial index and then created a new index based-on the model schema's definition.) We recognise this is not a supported Loopback method of adding indexes; however, the underlying issue appears to affect **all** multi-column indexes. 

See below for more details for debugging the issue.

## Additional information
linux x64 12.17.0

+─ loopback@3.25.1                                                                                                                         
+─ loopback-boot@2.28.0                                                                                                                    
+─ loopback-component-explorer@2.7.0                                                                                                             
+─ loopback-component-migrate@0.0.0-semantically-released.0 (github:luminlife/loopback-component-migrate#e96bc2ea144b2d6221d82303c0edda96584a9469)
+─ loopback-connector-postgresql@5.2.0                                                                                                     
+─ loopback-ds-readonly-mixin@2.0.4                                                                                                        
+─ loopback-ds-timestamp-mixin@3.4.1                                                                                                       
+─ loopback-jwt@1.0.2 (github:luminlife/loopback-jwt#5724ef89725aeb1a9b9704ce7d4ba023bc575913)   

The code causing the issue appears to be here: https://github.com/strongloop/loopback-connector-postgresql/blob/ffb08f4ca60a711d544ce13996afebcf09b86712/lib/migration.js#L826

`si.keys` is an array of arrays built in the `normalizeIndexDefintion()` method of `migration.js` (e.g. `[["firstName", "ASC"],["lastName", "ASC"]]`), so,  given how `Object.keys` operates on arrays, `Object.keys(si.keys)` will be an array of the position indexes of each of the array property pairs (e.g. `siKeys = [0,1]`); however, when the comparison for matching existing database indexes is performed (on line 612 of `migration.js`):
```javascript
siKeys.forEach(function(propName, iter) {                                
  identical = identical && self.column(model, propName) === i.keys[iter];
});                                                                      
```
you'll notice the elements of `siKeys` are expected to be the property names (`propName`) of the keys (e.g. instead of the array of position index values `[0,1]`, `siKeys` is expected to be `["firstName", "lastName"]`), so the database index being evaluated will never match to any definition on the Loopback model and, therefore, the index will be dropped and recreated. 

It looks like the correct code for defining `siKeys` should be something like:
```javascript
const siKeys = si.keys.map((key) => {
   return key[0]; // get the property name described by "key"
});
```
this will create the correct array of property names and allow them to be matched to the column names on the model.

ong-james:
Multicolumn indexes are always being dropped on autoupdate regardless of if
their definition has changed. Changes:

- Fix mapping of property name from the keys of normalized index definition
- Add multicolumn index tests

Fixes #464 

## Checklist

### DIFF
--- a/test/postgresql.autoupdate.test.js
+++ b/test/postgresql.autoupdate.test.js
@@ -4,6 +4,8 @@
 // License text available at https://opensource.org/licenses/Artistic-2.0
 
 'use strict';
+const async = require('async');
+const sinon = require('sinon');
 const assert = require('assert');
 const _ = require('lodash');
 let ds, properties, SimpleEmployee, Emp1, Emp2;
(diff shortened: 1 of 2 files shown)


ong-james:
When a signed the commit, I'd forgot to switch my git user config to my work GitHub user account (james@stagehand.app) and instead signed with a personal account (j.simon.ong@gmail.com). I don't think there's a way to undo this, but if it's an issue I can submit a new PR with the correct commit author.

agnes512 (approved):

agnes512 (changes requested):
Sorry that I accidentally approved it. I will schedule some time to review it next week.

test/postgresql.autoupdate.test.js
```diff
@@ -677,4 +679,274 @@ describe('autoupdate', function() {
...
+      it('removes existing index successfully', (done) => {
+        const noIndexSchema = {
+          ...schema,
+          options: {
+            ...schema.options,
+            indexes: {},
+          },
+        };
+
+        async.series([
```
bajtos, line 757: Let's rewrite this test from callback style to `await` style please.

```js
it('removes existing index successfully', async () => {
  // ...
  await ds.autoupdate([schema.name]);
  // ... 
  const indexes = ds.connector.discoverModelIndexes(schema.name);
  // ...
});
```

If the method `discoverModelIndexes` does not support promises yet, then you can add support for promises. See https://github.com/strongloop/loopback-connector-couchdb2/pull/83 for inspiration, I think we don't need a dedicated test for this change, since the method is not considered as a public API. Alternatively, you can add a small wrapper to this test file:

```js
function discoverModelIndexes(ds, schemaName) {
  return new Promise((resolve, reject) => {
    ds.connector.discoverModelIndexes(schemaName, (err, indexes) => {
       if (err) reject(err) else resolve(indexes);
    });
}
```

devops-juxbox: @bajtos Yes, I'll make the change. `disoverModelIndexes` doesn't support promises, so that's why I was using the callback.

ong-james: @bajtos Yes, I'll make that change to add support for promises. `discoverModelIndexes` doesn't yet support promises, so that's why I was using callbacks.

test/postgresql.autoupdate.test.js
```diff
@@ -677,4 +679,274 @@ describe('autoupdate', function() {
...
+      await personModel.destroyAll();
+    });
+
+    describe('when table has no existing indexes', () => {
+      it('updates without errors', async () => {
+        personModel = ds.define(schema.name, schema.properties, schema.options);
+        await ds.autoupdate([schema.name]);
+        assert(true, true);
+      });
+    });
```
agnes512, line 736: I am a bit confused here. Does this test mean that when we first create the table and it doesn't have existing indexes in the DB?

test/postgresql.autoupdate.test.js
```diff
@@ -677,4 +679,274 @@ describe('autoupdate', function() {
...
+            },
+          };
+
+          it('drops the existing index', async () => {
+            ds.define(updatedSchema.name, updatedSchema.properties, updatedSchema.options);
+            await ds.autoupdate([updatedSchema.name]);
+
+            const calls = spy.getCalls();
+            const match = calls.some(call => DROP_INDEX_REGEX.test(call.firstArg));
+            assert(match);
```
agnes512, line 850: I think it'd be better to add ` ds.connector.discoverModelIndexes(schema.name...)` so that we can see the result of this update.

test/postgresql.autoupdate.test.js
```diff
@@ -677,4 +679,274 @@ describe('autoupdate', function() {
...
```
agnes512, line 874: ditto

test/postgresql.autoupdate.test.js
```diff
@@ -677,4 +679,274 @@ describe('autoupdate', function() {
...
```
agnes512, line 898: ditto

agnes512 (commented):
Thanks for the PR! I have a few comments about the tests.

test/postgresql.autoupdate.test.js
ong-james: @agnes512 Essentially this is just performing the same test as the single column index test here: https://github.com/strongloop/loopback-connector-postgresql/blob/09791c6ec7b502d285f6a70202eee91dd92e9092/test/postgresql.autoupdate.test.js#L662
I could add assertions to both tests to further validate that the indexes exist on the models following the `autoupdate()` calls (or at least add it to my test.) The intent of the test as-is, though, is just to check that no errors are thrown.

test/postgresql.autoupdate.test.js
ong-james: @agnes512 I can certainly add the validation against the `ds.connector.discoverModelIndexes()` result (and to the tests referenced below, as well.) Just to be clear, are you suggesting removing the check for the `DROP` SQL calls and replacing with _just_ the check against the modified schema? That should be fine since if the schema was changed, then the `DROP` must have happened.

slnode:

PR #498, opened 2022-04-18
fix: debug prints the password in plain text

frbuceta:
Signed-off-by: Francisco Buceta <frbuceta@gmail.com>

## Checklist

- New tests added or existing tests modified to cover all changes
- Code conforms with the [style guide](https://loopback.io/doc/en/contrib/style-guide-es6.html)

### DIFF
--- a/lib/postgresql.js
+++ b/lib/postgresql.js
@@ -92,7 +92,17 @@ function PostgreSQL(postgresql, settings) {
   }
 
   this.settings = settings;
-  debug('Settings %j', settings);
+  debug('Settings %j', {
+    ...settings,
+    ...(typeof settings.url !== 'undefined' && {
+      get url() {
+        const url = new URL(settings.url);
+        if (url.password !== '') url.password = '***';
+        return url.toString();
+      },
+    }),
+    ...(typeof settings.password !== 'undefined' && {password: '***'}),
+  });
 }
 
 // Inherit from loopback-datasource-juggler BaseSQL


dhmlau (approved):

dhmlau:
@frbuceta, the changes LGTM. I was trying to kick off CI again for the failed tests (`build (12)`) but now all failed. :(

frbuceta:
> @frbuceta, the changes LGTM. I was trying to kick off CI again for the failed tests (`build (12)`) but now all failed. :(

The tests in the main branch are failing... I'm going to create a new PR with the changes I've made in the workflow

frbuceta:
See #499

lib/postgresql.js
```diff
@@ -92,7 +92,10 @@ function PostgreSQL(postgresql, settings) {
   }
 
   this.settings = settings;
-  debug('Settings %j', settings);
+  debug('Settings %j', {
+    ...settings,
+    ...(typeof settings.password !== 'undefined' && {password: 'HIDDEN'}),
```
achrinza, line 97: suggestion: Take into consideration `url` and use asterisks instead, which is more universally recognisable and language-agnostic.

```suggestion
    ...(typeof settings.password !== 'undefined' && {url: settings.url.replace(settings.password, '***') password: '***'}),
```

achrinza (changes requested):

lib/postgresql.js
frbuceta: If a URL is defined, it is not necessary to define a password outside of it. Therefore this would not cover it 100%. (See my last change on the commit)

frbuceta: Example:
Url Connection: _postgresql://other:**secret**@localhost/otherdb?connect_timeout=10&application_name=myapp_
After Url Connection: _postgresql://other:***@localhost/otherdb?connect_timeout=10&application_name=myapp_

achrinza: You're right. Initially I though Juggler's `DataSource` would populate `settings` in both of these syntax:

```javascript
const {DataSource} = require('loopack-datasource-juggler');
// "Single URL" syntax
new DataSource('mssql://user:pass@127.0.0.1');
// "`url` settings" syntax
new DataSource({url: 'mssql://user:pass@127.0.0.1');
```

But apparently the "`url` settings" syntax does not cause the other settings fields to be populated (e.g.`password`), only the "Single URL" syntax.

achrinza (approved):
The changes LGTM :+1:

PR #499, opened 2022-04-18
ci: node npm version v12 and v14 cannot process lock file v2

frbuceta:
Signed-off-by: Francisco Buceta <frbuceta@gmail.com>

## Checklist

- `npm test` passes on your machine
- New tests added or existing tests modified to cover all changes


.github/workflows/continuous-integration.yml
```diff
@@ -21,28 +24,32 @@ jobs:
...
       - name: Use Node.js ${{ matrix.node-version }}
-        uses: actions/setup-node@v2
+        uses: actions/setup-node@v3
         with:
           node-version: ${{ matrix.node-version }}
-      - run: npm install
-      - run: npm test
+      - name: Update NPM
+        if: matrix.node-version <= 14
+        run: npm install -g npm@latest-7
```
achrinza, line 34: From my testing, updating NPM may not be necessary.

.github/workflows/continuous-integration.yml
```diff
@@ -21,28 +24,32 @@ jobs:
...
+      - name: Install dependencies
+        run: npm ci
```
achrinza, line 36: if `--ignore-scripts` works, we should keep it there. This can be used to signal that this package works without the `postinstall` scripts executing (i.e. for the more security-conscious users).

.github/workflows/continuous-integration.yml
```diff
@@ -21,28 +24,32 @@ jobs:
...
+      - uses: actions/checkout@v3
+      - uses: actions/setup-node@v3
         with:
-          node-version: 14
-      - name: Bootstrap project
-        run: |
-          npm ci --ignore-scripts
+          node-version: 16 # LTS
+      - name: Install dependencies
+        run: npm ci
```
achrinza, line 52: Ditto
```suggestion
        run: npm ci --ignore-scripts
```

.github/workflows/continuous-integration.yml
```diff
@@ -51,15 +58,13 @@ jobs:
...
-        uses: actions/setup-node@v2
+      - uses: actions/setup-node@v3
         with:
-          node-version: 14
-      - name: Bootstrap project
-        run: |
-          npm ci --ignore-scripts
+          node-version: 16 # LTS
+      - name: Install dependencies
+        run: npm ci
```
achrinza, line 68: Ditto
```suggestion
        run: npm ci --ignore-scripts
```

achrinza (commented):
I've checked the CI runs in https://github.com/loopbackio/loopback-connector-postgresql/pull/499 and it doesn't seem that `package-lock.json` v2 is the issue; It seems to fail at the tests instead.

`package-lock.json` v2 is an intermediate format that combines both v1 and v3, so it should be supported by all NPM versions:

https://github.com/npm/rfcs/blob/main/implemented/0043-lockfile-version-config.md

.github/workflows/continuous-integration.yml
frbuceta: Github actions is currently overwriting the package-lock file during testing. If you change and do npm ci will fail

```
npm WARN read-shrinkwrap This version of npm is compatible with lockfileVersion@1, but package-lock.json was generated for lockfileVersion@2. I'll try to do my best with it!
```

frbuceta: The documentation says that NPM v7 supports package-lock v1, v2 and v3 but Node v12 and v14 come with lower NPM versions

achrinza: AFAIK, the warning is just a warning as they didn't update NPM <=v6 to explicitly recognise v2 lockfiles, but v2 lockfiles were designed to be backwards-compatible nonetheless.

From https://blog.npmjs.org/post/626173315965468672/npm-v7-series-beta-release-and-semver-major.html:

> package-lock.json is updated to a newer format, using "lockfileVersion": 2. This format is backwards-compatible with npm CLI versions using "lockfileVersion": 1, but older npm clients will print a warning about the version mismatch

Hence my understanding of the NPM-lockfile support is:

- NPM <=v6
    - Can read v1 and v2 lockfiles, but not v3
    - Can generate v1 lockfiles, but not v2 (will gracefully downgrade) or v3 (will regenerate from scratch)
- NPM v7 & v8
    - Can read v1, v2 and v3 lockfiles
    - Can generate v2 (default) and v3 lockfiles, but not v1 (will gracefully upgrade).

> Github actions is currently overwriting the package-lock file during testing

The switch to `npm ci` solves this problem as NPM will not try to modify the lockfile. Hence we don't need to upgrade the NPM version.

.github/workflows/continuous-integration.yml
```diff
@@ -0,0 +1,71 @@
...
+        with:
+          node-version: ${{ matrix.node-version }}
+      - name: Install PostgreSQL Tools
+        run: sudo apt install postgresql-client
+      - name: Install dependencies
+        run: npm ci --ignore-scripts
+      - name: Load PostgreSQL schema
+        run: psql -h 127.0.0.1 -U postgres < ./test/schema.sql
+      - name: Run tests
+        run: npm test --ignore-scripts
```
achrinza, line 39: As per-https://github.com/loopbackio/loopback-connector-postgresql/pull/499#discussion_r852447000, we can't reliably use `npm test --ignore-scripts` or `npm run <script name> --ignore-scripts` due to inconsistencies across NPM versions. We should create a `test:ci` script (which would not have any lifecycle scripts) and use that instead:

```suggestion
        run: npm run test:ci
```

achrinza (changes requested):

achrinza:
Closing as part of general cleanup, and for the reasons stated above.

PR #506, opened 2022-09-16
Fix ORDER BY logic

ataft:
This should be `if(orderBy)` rather than `if(!orderBy)`


samarpanB (approved):

samarpanB:
@ataft DCO check is failing. Can you please sign the commit ?

PR #781, opened 2025-06-27
Fix: Connection terminated unexpectedly error not caught in on error listener

rchandranmani:
Our application was crashing when the database connection unexpectedly failed. This happened because:

1. We were previously using `removeAllListeners('error')`, which unintentionally removed the pg library's error handlers.
2. This left connection termination errors unhandled, leading to app crashes which uses this connector

We've fixed this by:

1. Completely removing the `removeAllListeners('error')` call within the connector. This ensures that the pg library's built-in error handling for connection issues remains active.
 2. We now only add our connector's specific error listener if the pg client doesn't already have any error listeners attached `(!pg.listenerCount('error'))`. This prevents us from creating too many listeners (addressing past MaxEventListeners warnings) while ensuring all critical connection errors are still properly managed.

Fixes https://github.com/loopbackio/loopback-connector-postgresql/issues/778
See also https://github.com/loopbackio/loopback-connector-postgresql/issues/746

## Checklist


aaqilniz:
Hi, @rchandranmani. Can you please fix the linting issues and the commit message?

You'd also need to add some tests around the changes you've made.

